PR #40, opened 2015-03-23
Handle an empty GIT repository for versioning

xhochy:
In the case where a pyscaffold project is moved into a new, yet empty GIT repository, git describe fails with exit code 128 due to not having any revision, i.e. there is also no HEAD. Only in this case handle the error and do the same actions as if an invalid tag (format) was given.

### DIFF
--- a/pyscaffold/data/setup_py.template
+++ b/pyscaffold/data/setup_py.template
@@ -211,6 +211,17 @@ def version_from_git(tag_prefix, root, verbose=False):
         if verbose:
             print("no git found")
         return None
+    # Check if we have a valid tag in the current git repository
+    try:
+        next(git("show-ref", "--tags"))
+    except subprocess.CalledProcessError as e:
+        if e.returncode == 1:
+            # git-show-ref returns with exit code 1 in the case of no commit.
+            # In the case of any other error this is != 1, mostly 128.
+            print("Empty repository, please make an initial commit.")
+            return None
+        else:
+            raise
     tag = next(git("describe", "--tags", "--dirty", "--always"))
     if not tag.startswith(tag_prefix):
         if verbose:


FlorianWilhelm:
If exit code 128 is used to exactly state this then wouldn't it be more robust to check the exit state instead of string comparison with the output?

xhochy:
No, exit code 128 seems to be the general code for "something went wrong". 

Also I just realised that that string comparison only works if your system language is English. To fix that, we could just possibly run this command with @LANG=C@ to force the English output but then if the error is not the above mentioned, it may not be in the user's native language. (Although we could probably assume that PyScaffold user have a sufficient understanding of English).

kopf:
Would it not be an idea to use proper git bindings, such as [GitPython](http://gitpython.readthedocs.org/en/latest/tutorial.html) throughout pyscaffold, instead of subprocess and output parsing (which is bound to break sooner or later)?

xhochy:
But that would introduce a dependency on GitPython in the setup.py or we would need to bundle it...both cases seem not so pleasant.

FlorianWilhelm:
@kopf I don´t think GitPython is a proper git binding as it seems to also just call the git binary and parse its output. Also, there is not much development either. PyScaffold was using GitPython up until version 0.1.2 and then I dropped it for these reasons.

kopf:
Ah, sorry, I'm unfamiliar with the python/git landscape (perhaps [pygit2](http://www.pygit2.org/index.html) would be more appropriate?) and simply wanted to make the point that perhaps using proper bindings would be a better way to go.

FlorianWilhelm:
@kopf thanks for pointing this out but I think this feature would lead to further restrictions to the user of PyScaffold. 
But back to the original topic!
@xhochy Couldn't we just check if there were any tags defined beforehand with a call to `git show-ref --tags`. I think a little function like `do_git_tags_exist()` which calls `git show-ref --tags` and returns True if there are returned lines and False if not, could help. What do you think?

xhochy:
@FlorianWilhelm sounds like a good alternative

FlorianWilhelm:
@xhochy Do you want to change this PR into it? Then I'll merge it. Don't forget to add yourself to the AUTHORS.txt file as well.

xhochy:
will do

FlorianWilhelm:
If `git show-ref --tags` also just returns zero lines if not in a git repository we could use `git rev-parse --git-dir` to make sure we are in a git repository.

xhochy:
I did some more research and found out that `git show-refs --tags``has different exit codes which we can use to distinguish the different situations thus we do not need to fallback to counting the number of ouput error lines.

FlorianWilhelm:
@xhochy Thanks a lot!

PR #53, opened 2015-05-13
tests dir should not contain __init__.py

HolgerPeters:
According to the pytest docs, the tests directory should not contain an __init__.py (then it is possible to run tests on other versions or releases of pyscaffold and test wheel releases which do not distribute the tests).

Instead a conftest.py is created. This file is used by pytest to share test fixtures, etc.


FlorianWilhelm:
Shouldn't we provide at least a template file for ``conftest.py``? Maybe a Docstring for the module that explains what the file is for and how one would extend it, maybe even with a link to the manual page of py.test?

FlorianWilhelm:
@HolgerPeters Thanks

Issue #76, opened 2015-10-09
issues with automatic version number


sebastianneubauer:
It seems, if the package setup.py is not in the git repository root directory, the generated version string is different than in the root dir:
root dir:
python setup.py --version
0.0.post0.dev1+11079f0
subfolder:
python setup.py --version
0.0.1.dev1

FlorianWilhelm:
This is kind of unintended usage. Did everything else work? Can you explain what you exactly did? 
Somehow setuptools_scm switches back to the default versioning.

sebastianneubauer:
Mmmh, I feared an answer like this, but in my experience we cannot come around accepting, that the pressure to use pyscaffold is sometimes smaller than the pressure to not have setup.py in the root of a repository. Often for example there are many packages in one repository, because they should have a common version (same version control)...then it makes sense to have them both in subfolders and both have a separate pyscaffold... 
As far as I can see, everything except for the versioning is working...of course in order to be really usefull and intended in this scenario, one should be able to tell pyscaffold not to create a repo...

FlorianWilhelm:
Ok, it should be no problem to let PyScaffold only create a repo if the current directory is not already a git repo. The problem is the strange versioning... do you want to look into it? My guess is it is setuptools_scm related. With a debugger it should be no problem to find the cause. I am on holidays right now and cant do it.

sebastianneubauer:
Ok, I can have a look...

sebastianneubauer:
I analyzed the issue with the versioning and found the reason. Setuptools has a configuration parameter "root" which defaults to ".".
We can quite easily fix this issue by setting the correct root dir. The root folder can be identified with the following:

     repo_dir = subprocess.Popen(['git', 'rev-parse', '--show-toplevel'], stdout=subprocess.PIPE).communicate()[0].rstrip().decode('utf-8')

FlorianWilhelm:
Thanks, that was fast. Is this a parameter of setuptools or setuptools_scm? The latter I guess. I'll think with a few changes in integration.py we can resolve this issue. I'll look into it after my holidays.

FlorianWilhelm:
Thanks to @treebee this is fixed in release v2.4.3!

PR #78, opened 2015-10-20
Use 'git rev-parse --show-toplevel' to get same version in subfolders

treebee:
This uses @sebastianneubauer 's solution for #76 to address the automatic versioning in subfolders.

### DIFF
--- a/pyscaffold/utils.py
+++ b/pyscaffold/utils.py
@@ -9,6 +9,7 @@ import keyword
 import os
 import re
 import sys
+import subprocess
 from contextlib import contextmanager
 from operator import itemgetter
 from distutils.filelist import FileList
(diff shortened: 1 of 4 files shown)


FlorianWilhelm:
Thanks, I'll have a look into it on Monday.

FlorianWilhelm:
Hi @treebee, could you rebase onto tag v2.4.2 in order to push this out faster?

pyscaffold/utils.py
```diff
@@ -209,3 +210,13 @@ def check_setuptools_version():
         raise RuntimeError(
             "ERROR: Your setuptools version is too old (<12).\n"
             "Use `pip install -U setuptools` to upgrade.")
+
+
+def get_git_root():
```
FlorianWilhelm: Could we move this function into the module ``repo.py`` since it is related to the handling of the git repository. Also please take a look into ``shell.py``, this module provides already a ```git`` command. that means the ``get_git_root`` function could be reduced to something like ``return git('git', 'rev-parse', '--show-toplevel')`` if I got it right.

FlorianWilhelm:
@treebee , looks good! Thank you. I'll merge and fix the unit test, after that I'll release v2.4.3.

PR #83, opened 2015-11-05
Fix miniconda install path to not depend on upstream consistency

cpaulik:
- currently miniconda for Python2.x installs into `miniconda2`
  by default which breaks this script


cpaulik:
Oh I'm sorry, this was against master. I'll open a new one against develop.

PR #84, opened 2015-11-05
Fix travis build and travis_install template

cpaulik:
- currently miniconda for Python2.x installs into `miniconda2`
  by default which breaks this script


FlorianWilhelm:
Thank you! I'll merge. Right now there is another small problem in the Travis build. I hope I'll have some time on the weekend to look into this.

PR #85, opened 2015-11-14
update PATH to match Miniconda2 envs

ipashchenko:
Hi! 

When ``PATH`` in ``tests/travis_install.sh`` is set according to ``export PATH=$HOME/miniconda/bin:$PATH`` then Travis fail with error:

```
tests/travis_install.sh: line 24: conda: command not found
```

for builds with ``DISTRIB="conda"``.

Changing to ``export PATH=$HOME/miniconda2/bin:$PATH`` resolves the issue


ipashchenko:
Ooops... It's seems like this is me-specific error with ``miniconda2``

PR #86, opened 2015-11-21
Query effective rather than global git configuration

kynan:
Recent versions of git support [additional configuration files](https://git-scm.com/docs/git-config#FILES) e.g. `$XDG_CONFIG_HOME/git/config`. When setting `user.name` and `user.email` in this file (as I do), it is not reported by `git config --global`.

I don't really see a reason to explicitly query the *global* git configuration instead of the *effective* git configuration, which is what a plain `git config user.name` does.

### DIFF
--- a/pyscaffold/info.py
+++ b/pyscaffold/info.py
@@ -26,7 +26,7 @@ def username():
     :return: user's name as string
     """
     try:
-        user = next(shell.git("config", "--global", "--get", "user.name"))
+        user = next(shell.git("config", "--get", "user.name"))
         user = user.strip()
     except CalledProcessError:
         user = getpass.getuser()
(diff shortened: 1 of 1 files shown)


FlorianWilhelm:
@kynan Good point, thanks a lot. If you want to add yourself to `AUTHORS.rst`, I'll merge then this PR.

kynan:
Thanks for merging and sorry for the slow response! Do you want me to submit a separate pr for the AUTHORS file?

FlorianWilhelm:
I can also add you, no problem. What's the mail address you want to use?

kynan:
OK, thanks. Please use the email address from my commit.

PR #87, opened 2016-02-05
fixed description to run example installation, convert terminal argum…

WernerTim:
…ent to int

### DIFF
--- a/pyscaffold/templates/skeleton.template
+++ b/pyscaffold/templates/skeleton.template
@@ -6,9 +6,9 @@ console script. To run this script uncomment the following line in the
 entry_points section in setup.cfg:
 
     console_scripts =
-        hello_world = ${package}.module:function
-
-Then run `python setup.py install` which will install the command `hello_world`
+     fibonacci = my_project.skeleton:run
+     
+Then run `python setup.py install` which will install the command `fibonacci`
 inside your current environment.
 Besides console scripts, the header (i.e. until _logger...) of this file can
 also be used as template for Python modules.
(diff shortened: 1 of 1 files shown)


pyscaffold/templates/skeleton.template
```diff
@@ -67,7 +67,7 @@ def parse_args(args):
 
 def main(args):
     args = parse_args(args)
-    print("The {}-th Fibonacci number is {}".format(args.n, fib(args.n)))
+    print("The {}-th Fibonacci number is {}".format(args.n, fib(int(args.n))))
```
FlorianWilhelm: Instead of converting the parameter, couldn't we just add ``type=int`` to 

>      parser.add_argument(
         dest="n",
         help="n-th Fibonacci number",
         metavar="INT")

pyscaffold/templates/skeleton.template
```diff
@@ -6,9 +6,9 @@ console script. To run this script uncomment the following line in the
 entry_points section in setup.cfg:
 
     console_scripts =
-        hello_world = ${package}.module:function
-
-Then run `python setup.py install` which will install the command `hello_world`
+     fibonacci = my_project.skeleton:run
```
FlorianWilhelm: I think we still need ``${package}`` instead of ``my_project`` for this to work.

FlorianWilhelm:
@WernerTim Thank you for pointing out that bug. I made a few comments above. If you want you can add yourself also to the ``AUTHORS.rst`` file by pushing more commits to this PR.
Nice would also be a test that something like this does not happen again. Once could just call ``fibonacci 12`` in  ``system_test.sh``.

pyscaffold/templates/skeleton.template
WernerTim: Yes, we still need ${package}

pyscaffold/templates/skeleton.template
WernerTim: Adding type=int is probably the cleaner solution

PR #93, opened 2016-04-12
Adds --with-vagrant option

brianbruggeman:
Motivation:

I am often building packages for an environment (debian8) that does not match my local box (a mac).  Consequently, I generally add a virtual machine provisioner (vagrant) to my package.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -10,3 +10,4 @@ Developers
 * Florian Rathgeber <florian.rathgeber@gmail.com>
 * Eva Schmücker <email@evaschmuecker.de>
 * Tim Werner <tim.werner@blue-yonder.com>
+* Brian Bruggeman <brian.m.bruggeman@gmail.com>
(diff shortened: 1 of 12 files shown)


HolgerPeters:
I do like that this dispatches to `vagrant init` and does not provide a full template that we have to maintain in pyscaffold. On the other hand, executing vagrant init manually after `putup` is not difficult at all.  I.e. I do not see a huge benefit of using `putup --with-vagrant myproject` over `putup myproject && cd myproject && vagrant init`.
To summarize: I think vagrantfile generation should be decoupled from pyscaffold.

brianbruggeman:
You may have missed my template (see: https://github.com/blue-yonder/pyscaffold/pull/93/files#diff-715cc25ad498112a1b7a826167892f0a), which is more than the standard `vagrant init`. 

At least in my experience over the past 5 years using Vagrant, rarely do my team members know Ruby and Vagrant.

PR #94, opened 2016-04-13
Adds --with-vagrant option

brianbruggeman:
Motivation:

I am often building packages for an environment (debian8) that does not match my local box (a mac). Consequently, I generally add a virtual machine provisioner (vagrant) to my package.

It appears that the testing is failing due to a library not being available on the pandas build (ARM), which is unrelated to this specific change.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -10,3 +10,4 @@ Developers
 * Florian Rathgeber <florian.rathgeber@gmail.com>
 * Eva Schmücker <email@evaschmuecker.de>
 * Tim Werner <tim.werner@blue-yonder.com>
+* Brian Bruggeman <brian.m.bruggeman@gmail.com>
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,7 @@ Version 2.5.6, 2016-??-??
 
 - Prefix error message with ERROR:
 - Suffix of untagged commits changed from {version}-{hash} to {version}-n{hash}
+- Adds support for vagrant (--with-vagrant)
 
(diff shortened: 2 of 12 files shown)


HolgerPeters:
I still (even though I have now realized that this PR in fact would add an option to create a curated Vagrantfile) think that the feature is a bit too far out-of-scope. What PyScaffold does well is to integrate all those numerous tools of the Python ecosystem into a compelling default set, which would be challenging without PyScaffold. Adding a Vagrantfile to a repository is orthogonal to this problem. It is not especially difficult, `vagrant init` brings you very far already. I don't feel PyScaffold has a significant gain from such a feature, but once we have this flag in pyscaffold, we must support it indefinitely.

HolgerPeters:
I'll close this PR now. Thank you for taking the time to put together a contribution, even though I cannot merge it. You are of course welcome to contribute to PyScaffold again. Discussing your implementation idea in an issue first, could help to avoid situations like this, where we do not feel comfortable merging a contribution.

sebastianneubauer:
Even if it is closed already, I would support that decision: The success of pyscaffold is for a good part based on the minimal approach. The less opinionated decisions it takes and the simpler it is (less options), the more people can and will use it. I love vagrant and I use it extensively, but the given template would have fitted probably only to a small subset of artifacts, furthermore not all packages are in a separated vagrant box. So I agree, it is kind of orthogonal to creating a minimal standard python package...
Maybe you can contribute your work on a "better starting point" for a vagrant file somewhere else, maybe vagrant itself, or even a small vagrantscaffold tool...thanks nonetheless for your contribution!

PR #103, opened 2017-05-21
Suggestion and proof of concept to add extension points and improve Python API

abravalheri:
Sorry guys for the big pull request. In fact I was experimenting with the concept of extension points and trying to improve the Python API for PyScaffold, and somehow it ended up really far.

I have tried to refactor some details of the implementation, while keeping what I though was the main API (namely `get_default_opts` and `create_project`). A more strict separation of CLI and API was also the target, as well as a way to isolate the implementation of new features (or options) from the main core, therefore making it easier to contribute to the project development. 

The main idea is organize `create_project` in 3 steps:
- before hooks
- the project creation itself
- after hooks

Extensions can register hooks and also change the project structure that will be created.

As a result, I could implement almost all `--with-*` options (the exception is the namespace, although I have not tried, and it might be possible) built-in extensions, pretty much isolated from the remaining python modules.

A fair question is whether or not this organization is useful for PyScaffold. Well ... at least I can see some advantages for eventual contributions. Regarding the creation of extensions: some alternative software (and here I talk specifically about cookiecutter) are extremely customable and may already cover the main use cases. But I think this implementation incentives the composition of different pieces and thus follows a different (and useful) approach.

So dear maintainer folks, what do you think about it?
Please guys, share any feedback or ideas, I will happily discuss all the details.

Finally, I would like to say that I had some trouble trying to run the test suite on Travis with Python 2.7 (a crazy `[pbr] Generating ChangeLog ERROR:root:Error parsing` before even running pytest) and locally (for some reason I always get an error while importing cookiecutter, no matter how many times I install it in different virtualenvs), but if you find this idea interesting I can spent more time to make it run more smoothly.


FlorianWilhelm:
Hi @abravalheri, thanks for your work and PR! It really is massive and it's gonna take some time to review. I'll let you know more as soon as I can. Thanks again for putting so much work into it, it's greatly appreciated.

FlorianWilhelm:
Hi @abravalheri, sorry for the delay! I am impressed by your changes and how you make PyScaffold more easily to extend. That's why I checked back with @BY-jk and he agreed to add you as a collaborator so that you can directly work on your changes that will then result in a version 3.0 of PyScaffold. Would that work for you?

abravalheri:
Hi @FlorianWilhelm.
Yes this would definitely work. Thank you for your kind offer, warm reception in the community, and also for the wonderful project!

I have already taken a look into the Contributing Guide, however, I would like to first understand a little bit more what is the development process behind PyScaffold in order to not make a mess (for example, to work for v3.0, should I open a new branch or continue to work with feature-branches? Is there any workflow that you follow? Is there any policies for PyScaffold, like keeping dependencies at minimum?).

May I send you an email with these doubts, or should we keep discussing it in the topic? (Or maybe gitter is more appropriate ...)

FlorianWilhelm:
@abravalheri I added you as collaborator and created a branch `v3.x` for this PR. Welcome!

FlorianWilhelm:
Regarding the parsing error, it seems that this is caused by Sphinx version above 1.6 which added the typing package as additional dependency. I don't know why it behaves like that but as a workaround I pinned the Sphinx version to 1.5.6 for the time being.

abravalheri:
Thank you very much @FlorianWilhelm (and sorry for the little mess).
With your workaround, I was able to fix the problems with the other versions of Python and Travis finally is green again.
I have uploaded the changes to branch v3.x as recommended.

FlorianWilhelm:
@abravalheri thanks for merging directly, I'll close the PR then.

Issue #102, opened 2017-04-05
Support for cookiecutter is under-documented


brianthelion:
First, pyscaffold is great! Thanks!

Second, cookiecutter support is a great addition. Without it, there's no way to do organization-, project-, or framework-specific boilerplate reduction.

As of right now, though, the `putup --with-cookiecutter` experience is a little bit confusing. The extent of the `putup --help` is:
```
PyScaffold is a tool for easily putting up the scaffold of a Python project.

positional arguments:
  PROJECT               project name

optional arguments:
...
  --with-cookiecutter TEMPLATE
                        additionally apply a cookiecutter template
```
The narrative documentation in the [README.md](url) is:
```
Django & Cookiecutter
=====================

Create a `Django project <https://www.djangoproject.com/>`_ with the flag
``--with-django`` which is equivalent to
``django-admin.py startproject my_project`` enhanced by PyScaffold's features.

With the help of `Cookiecutter <https://cookiecutter.readthedocs.org/>`_ it
is possible to customize your project setup. Just use the flag
``--with-cookiecutter TEMPLATE`` to use a cookiecutter template which will be
refined by PyScaffold afterwards.
```
For the naive user, there is (A) a tendency to want to believe that the `--with-cookiecutter` option can be applied to *any existing* cookiecutter template when, in fact, (B) the template needs to be designed to work with pyscaffold. Unintended consequences are par for the course.

To combat #A, there should probably be a warning in both the README and the `--help`.

To combat #B, there should be some narrative documentation that explains how to build a cookiecutter template for pyscaffold.

FlorianWilhelm:
Thanks for your input @brianthelion. I totally agree. The story of the cookiecutter feature is actually that it was requested by a former colleague, implemented and I don't know if ever really used. It is actually quite simple, PyScaffold calls cookiecutter with all parameters (author name, e-mail, package name etc.) and later does its actual job. I haven't myself used this feature so far in practice. Maybe someone from Blue Yonder can comment on that. 
Are you using this feature? It would be great if you could contribute the missing documentation.

brianthelion:
@FlorianWilhelm I would like to use the feature, but will need to author new templates to accomplish what I need to as in #B above. That reduces the overall utility of pyscaffold for me, as there are already cookiecutter templates that do what I want and won't require modification. This sets up an "pyscaffold vs cookiecutter" scenario that I am guessing plenty of other folks are dealing with as well.

[...]

PR #109, opened 2017-09-01
Extension Points and Embeddable APIs

abravalheri:
This pull request is the continuation of the work of #103 and a sort of Request for Comments/Review.

As discussed with @FlorianWilhelm, the previous API was not flexible enough to determine how the extension hooks would interact with the execution flow.

This new implementation defines the concept of a pyscaffold `action` that is a well defined, single-purposed and incremental change in the "master-plan" for creating the project directory. It also enables extensions to insert new actions in any moment of the execution flow (except during the bootstrap period).

Additionally, this change introduces:
- a `list-actions` option to assist extensions development
- a `dry-run` option
- a new log system (more verbose, since it should be compatible with the existence o a `dry-run` option)
- better docs for using cookiecutters (related to #102)
- improved `with-namespace` option (also related to #102)
   - indeed, the namespace feature was also extracted into a separated extension, as a proof of concept of the power of the new extension API.


FlorianWilhelm:
@abravalheri Wow, you have been busy it seems :-) I hope to have enough time over the weekend to look into this but judging from your mail it looks really good.

abravalheri:
@FlorianWilhelm sorry for the extra work :sweat_smile: 

Maybe starting with the new docs would help? It is a quick way to get a general overview, check if everything makes sense, and if does not, abbreviate the review process...

Thanks a lot for the support anyway.

pyscaffold/termui.py
```diff
@@ -0,0 +1,87 @@
...
+    on_green=42,
+    on_yellow=43,
+    on_blue=44,
+    on_magenta=45,
+    on_cyan=46,
+    on_white=47
+)
+
+
+def isatty(stream=None):
```
FlorianWilhelm: Incomplete docstring with Args: and Returns:

docs/extensions.rst
```diff
@@ -4,49 +4,215 @@
...
+
+.. code-block:: python
+
+    def extend_scaffold(actions, helpers):
+        """Register custom actions that apply extension behavior."""
+
+        # ... define `start_hook`, `finish_hook` and `define_awesome_files`
+        #     functions
+
+        actions = helpers.register(actions, start_hook, before='create_structure')
```
FlorianWilhelm: Great explanation! Should it mean rather `before='define_structure'` here? Wouldn't that be more aligned with the more detailed explanation below?

pyscaffold/api/helpers.py
```diff
@@ -0,0 +1,372 @@
...
+                ['docs', 'api', 'index.html']
+        content (str): file text contents
+        update_rule: see :class:`~.FileOp`, ``None`` by default
+
+    Note:
+        Use an empty string as content to ensure a file is created empty.
+    """
+    # Ensure path is a list.
+    if isinstance(path, string_types):
+        path = path.split('/')
```
FlorianWilhelm: Using '/' here is according to your documentation but could lead to user errors on windows. `os.sep` holds the right path separator. One could even think about using `https://docs.python.org/3/library/pathlib.html` (and it's Python2 backport) that allows a proper handling of paths.

pyscaffold/api/helpers.py
```diff
@@ -0,0 +1,372 @@
...
+
+def ensure(structure, path, content=None, update_rule=None):
+    """Ensure a file exists in the representation of the project tree
+    with the provided content.
+    All the parent directories are automatically created.
+
+    Args:
+        structure (dict): project representation as (possibly) nested
+            :obj:`dict`. See :obj:`~.merge`.
+        path (str or list): file path relative to the structure root.
```
FlorianWilhelm: My experience is that functions that have arguments allowing varying data types end up to be complicated and are hard to test. One could think about forcing the user to provide an actual path as string (or `pathlib.Path` object) and telling him in the doc that he/she can easily use `os.path.join` to transform a list of directories into a proper path string (or with the corresponding `pathlib` function in the latter case).

docs/extensions.rst
```diff
@@ -4,49 +4,215 @@
...
+The :obj:`~pyscaffold.api.helpers.unregister` function accepts as second
+argument a position reference which can similarly present the module name::
+
+        helpers.unregister(actions, 'init_git')
+        helpers.unregister(actions, 'pyscaffold.api:init_git')
+
+.. note::
+
+    These functions **DO NOT** modify the actions list, instead they return a
+    new list with the changhes applied.
```
FlorianWilhelm: Typo: changes

docs/extensions.rst
```diff
@@ -64,62 +230,121 @@ The following example illustrates a simple extension implementation:
...
-
-Note that both :attr:`~pyscaffold.api.Scaffold.before_generate` and
-:attr:`~pyscaffold.api.Scaffold.after_generate` hooks also should be
-defined as a function of a single argument, a
-:class:`~pyscffold.api.Scaffold` instance.
+            })
+
+            structure['.python-version'] = ('3.6.1', helpers.NO_OVERWRITE)
+
+            for filename in opts['awesome_files']:
```
FlorianWilhelm: Cool examples but at that point it might not be clear to the reader how to get `awesome_files` entry into the `opts` dictionary. One would need to do this somehow in a function that was registered before and adds `awesome_files` to the options, right?

docs/extensions.rst
```diff
@@ -191,3 +416,40 @@ and can be used as reference implementation:
...
+
+When writing extensions, it is important to be consistent with the default
+PyScaffold behavior. In particular, PyScaffold uses a ``pretend`` option to
+indicate when the actions should not run but instead just indicate the
+expected results to the user, that **MUST** be respected.
+
+The ``pretend`` option is automatically observed for files registered in
+the project structure representation, but complex actions may require
+specialized coding. The :mod:`~pyscaffold.api.helpers` argument provides a
+special :class:`logger <pyscaffold.log.ReportLogger>` object usefull in
```
FlorianWilhelm: Typo: useful

docs/extensions.rst
```diff
@@ -64,62 +230,121 @@ The following example illustrates a simple extension implementation:
...
+    the correct location of the files relative to the current working
+    directory.
+
+As shown by the previous example, the :mod:`~pyscaffold.api.helpers` argument
+also presents constants that can be used as metadata. The ``NO_OVERWRITE`` flag
+avoids an existing file to be overwritten when ``putup`` is used in update
+mode.  Similarly, ``NO_CREATE`` avoids creating a file from template in update
+mode, even if it does not exist.
+
+.. note::
```
FlorianWilhelm: Well at that point your remark about `testability` and injection in your mail came back to my mind. Injection the helper module seems to actually complicate things here and as you wrote telling the user to `from pyscaffold import helpers` might be much more natural since not everyone feels comfortable dealing with `functools` or understanding closures in the former case. 
As always it's a tradeoff, is there anything we are winning by doing the injection of `helpers` compared to doing an `from pyscaffold import helpers` inside except of saving the user to write one line? I am not completely sure but in case there is no other reason at least my gut feeling says that the import is less complicated and convoluted.

docs/extensions.rst
```diff
@@ -145,10 +370,10 @@ function may be implemented:
         parser.add_argument('--with-awesome',
                             dest='extensions',
                             action='append_const',
-                            const=extend_pyscaffold,
+                            const=extend_scaffold,
                             help='generate awesome extra files')
 
-Note that, in this case, an option with the ``append_const`` action is created,
+In this case, an option with the ``append_const`` action is created,
 with ``extensions`` as ``dest`` and the extension function as ``const``.
```
FlorianWilhelm: If I got it right, the `extensions` destination is our special convention here that holds all activated extensions and the PyScaffold code will use the `extensions` attribute from the argparse namespace to look them up. If this is right maybe it makes sense to clarify this with a sentence or two at that point.

docs/extensions.rst
abravalheri: Do you think it is easier to understand? How about `after='create_structure'` should I keep it or change to `'define_structure'` as well?

pyscaffold/api/helpers.py
abravalheri: Do you think that it is ok to use the `pathlib` for the internal implementation and add the backport for Python 2 as a requirement for PyScaffold?

FlorianWilhelm: Yes, I think that is okay. Most people will start projects that are Python3-only anyway these days, I guess.

docs/extensions.rst
FlorianWilhelm: Ahh.. maybe that's just my mistake. I was comparing it to the later defined 
```python
helpers.register(actions, hook1, before='define_structure')
helpers.register(actions, hook2, after='pyscaffold.structure:create_structure')
```
and was confused. At that point I mixed something up. So it's okay from my point.

PR #111, opened 2017-09-03
Add support for GitLab-CI

smartsammler:
* A new flag `--with-gitlab-ci` lets PyScaffold create a
``.gitlab-ci.yml`` as an example for `GitLab <https://gitlab.com>`_ CI's configuration.

* The ``.gitlab-ci.yml``-template is based on the official ruby template, since there was no Python template in the GitLab instances I am using.
* The implementation is similar to `travis.yml`'s implementation

PR #113, opened 2017-09-09
Fix failing coverage and doc builds in the gitlab-ci

smartsammler:
I'm sorry I figured out this errors only now when testing your, @FlorianWilhelm, RC1 `2.5.8rc1`. Now everythink should be fine.
Thank you for merging #111 already into the version 2.

Fixes:
* Coverage is run at every test anyways so there is no need for an extra coverage run
* `Sphinx` needed to be installed first, since it is not included in the `test-requirements.txt`

### DIFF
--- a/pyscaffold/templates/gitlab_ci.template
+++ b/pyscaffold/templates/gitlab_ci.template
@@ -56,11 +56,9 @@ py36:
   script:
   - python setup.py test
 
-cov:
-  script:
-  - py.test --cov=cassandra tests
-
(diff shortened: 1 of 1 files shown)


FlorianWilhelm:
@smartsammler Thanks for the fast fix!

FlorianWilhelm:
@smartsammler I uploaded a version `2.5.8rc2` just to be 100% sure before I push a final `2.5.8`.

PR #121, opened 2017-10-08
Fix typo in extensions.rst

smartsammler:
changhes -> changes


FlorianWilhelm:
Hi @smartsammler, thanks for reporting this typo. I fixed it directly in the master since this was easier in that case.

PR #122, opened 2017-10-08
Make docu (extensions.rst) stick to convention

smartsammler:
The `Conventsions for Community Extensions` say you should name your extension `pyscaffold`*ext*`.extensionname`, but the documentation uses `pyscaffold.contrib`. If I got it right `contrib` is the extension's name (`$EXTENSION`) and not PEP423's `contrib` which is supposed to be replaced with `ext`. Therefore I changed
`pyscaffold.contrib` to `pyscaffoldext.contrib`.


FlorianWilhelm:
Sounds good to me.

Issue #116, opened 2017-09-11
Consider dropping PBR


FlorianWilhelm:
Newer versions of setuptools allow configuration with the help of `setup.cfg` for almost all options. A few years ago this was not possible and therefore PBR was used to achieve this. PBR itself is considered by many people as quite a hack and it is really hard to debug sometimes. For version 3.0 of PyScaffold we could consider dropping PBR and replacing it with the standard functionality of more recent versions of setuptools. Checkout the [setuptools' documentation](http://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files) for more details.

HolgerPeters:
PBR was the reason I stopped actively using pyscaffold.

abravalheri:
+1 for sure.
But let me ask you what would be the impact of this change on issue #48?
The idea would be allow the user to directly modify setup.py, or would
PyScaffold stick with the declarative approach using setup.cfg?

Do this change also affects #79?

>

FlorianWilhelm:
Yes, this would fix #48 since Cython usage would be done just as it is meant to be according to setuptools by changing `setup.py` directly. Most other more common configuration would still be done via `setup.cfg` in a declarative way (see [here](http://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files)) because it is just way easier and one of the more important reasons people like PyScaffold.
This would also affect #79. Currently, PBR reads in `requirements.txt` to generate to package dependencies. After we switched back to pure setuptools this link would be broken, meaning that one has to configure dependencies in setup.cfg and can define whatever is wanted in `requirements.txt`. I think this is okay since many people want to define dependencies without version pinning for the actual package but with version version pinning in requirements.txt. One could even provide a PyScaffold plugin/extension allowing for a `requirements.in` to be used with `pip-compile` and `pip-sync` as described [here](http://nvie.com/posts/better-package-management/). All these are things I have been asked for but could not include due to the dependency to PBR and the fact that I did not want to trade the declarative `setup.cfg` for another feature. So now with newer versions of setuptools, this is possible. 
Also #98 could be solved by the switch since PBR also handles the call to [`sphinx-apidoc`](http://www.sphinx-doc.org/en/stable/invocation.html#invocation-apidoc) right now which we would then do allowing us to fix this issue but of course this also comes with cost of maintaining this feature.

abravalheri:
[...]

Issue #115, opened 2017-09-11
Consider using `src` as source folder


FlorianWilhelm:
Currently PyScaffold creates a structure `my_project/my_package`. According to more recent versions of [setuptools](http://setuptools.readthedocs.io/en/latest/setuptools.html) a structure like `my_project/src/my_package` seems to be the preferred way.
Since we are preparing a major version 3.0 for PyScaffold we could also opt to change the current structure to the recommended one and consider it a breaking change. 
Any thoughts on this?

abravalheri:
There are a few thoughts about it in the following post blog:

https://blog.ionelmc.ro/2014/05/25/python-packaging/

Although it is a little bit old, the selling points seems quite fair.

There is also a companion template repository in
https://github.com/ionelmc/python-nameless with a few things that can
inspire PyScaffold.

FlorianWilhelm:
This is now in the current master for testing.

Issue #98, opened 2016-12-02
not updating newly added modules and remove removed ones


enfeizhan:
I am currently using pyscaffold 2.5.7 and sphinx 1.4.9.  Found two maybe "issues", firstly documentations for removed modules still exist e.g. in html files. It doesn't find the changes also the other way, newly added modules don't show up in htmls.

Wondering is this supposed to be so or I am not doing it correctly? Let me know if you need more info to identify any issue. Thanks!

enfeizhan:
Or this is actually a sphinx issue?

enfeizhan:
Also is there any way to ignore some module like the versioneer module?

FlorianWilhelm:
Hi @enfeizhan, thanks for reporting. I can confirm your two issues. PyScaffold aktuelle uses PBR for this functionality. A workaround is to delete ``docs/api`` before rebuilding the docs. Regarding your exclusion of files question have a look here: http://docs.openstack.org/developer/pbr/#pbr

FlorianWilhelm:
Could you try activating ``autodoc_index_modules = True`` in your ``setup.cfg`` and see if that helps with new module files? In this case PBR will itself generate the api docs.

FlorianWilhelm:
I added following bug report in PBR https://bugs.launchpad.net/pbr/+bug/1647140

enfeizhan:
Hi @FlorianWilhelm,
   Thanks a lot for the detailed reply. I tried your recommendations. It seems only deleting ``docs/api`` works.

  What I just did was, (I have found another workaround by manually adding or removing items in ``docs/api/myproject.rst``) first I turned ``autodoc_index_modules`` on and did ``python setup.py docs`` again. I expected I would see the modules I have manually removed from ``docs/api/myproject.rst`` appear again, but they didn't.
  Then I removed ``docs/api`` completely, this time the then removed module did appear.
   Lastly I tried to exclude them from ``autodoc_exclude_modules`` as in the openstack tutorial. Unfortunately the syntax is unclear to me so I am not sure if I have done it correctly so the modules I want to exclude are still there.

FlorianWilhelm:
Hi @enfeizhan, I created a project ``test_proj`` with files `new_module1-3.py` including in ``setup.cfg``:
```
[pbr]
# Let pbr run sphinx-apidoc
autodoc_tree_index_modules = True
autodoc_tree_excludes = test_proj/new_module2.py test_proj/new_module3.py
[...]

Issue #92, opened 2016-03-09
Cannot build or install project with git submodule


willu47:
When `git submodule` is used to create a sub-repository within a repo, `python setup.py install sdist bdist` fail with the error: 
    `error: can't copy 'exactVerify/shipquant': doesn't exist or not a regular file`

FlorianWilhelm:
This is most likely a bug in PBR as pointed out in the [PBR bugtracker](https://bugs.launchpad.net/pbr/+bug/1530867). When it is fixed in PBR I will update the version PyScaffold uses. My recommendation is to use ``git subtree`` anyway. You can check out how I use it in PyScaffold for PBR and setuptools-scm.

FlorianWilhelm:
@willu47 The problem still exists in version 2.5.6 since the authors of PBR seem to ignore this issue right now. Another workaround is to use either ``python setup.py develop`` or ``python setup.py sdist``  followed by ``pip install dist/your_package.tar.gz``. It seems only ``python setup.py install`` is choking on the submodule directory.

willu47:
Okay, thanks for the update.  We've found a workaround for now.

FlorianWilhelm:
Since PBR will be dropped with version 3 of PyScaffold, this bug is no longer present in the current master branch.

PR #125, opened 2017-10-14
Drop pbr

FlorianWilhelm:
This PR removes [pbr](https://docs.openstack.org/pbr/latest/) from PyScaffold. Starting from PyScaffold version 2.5 pbr was used to allow the configuration of a project with the help of `setupt.cfg`. Since setuptools version 30.3.0 this is now directly supported in an official way giving means for dropping PBR which is rather hacky anyway. The only problem right now is that there is a [bug](https://github.com/pypa/setuptools/issues/1136) causing setuptools to break when reading `setup.cfg` for Python2. This will surely be fixed and most people switched to Python3 anyway.
This PR addresses issue #116, #115, #98, #79, #92 and #48.
Some polishing still needs to be done like updating all docs etc. but since this is a major change it makes sense to first resolve the technical aspects.


FlorianWilhelm:
@abravalheri, just wanted to check in :-) Are you fine with the code? I would like to merge it in about one week to go on with the preparations for a version 3.

PR #127, opened 2017-11-02
Update __init__.template

smartsammler:
Add the exception name to the except clause since it is is considered best practice [(PEP-8:](https://www.python.org/dev/peps/pep-0008/#id51) "When catching exceptions, mention specific exceptions whenever possible instead of using a bare except: clause.")

### DIFF
--- a/pyscaffold/templates/__init__.template
+++ b/pyscaffold/templates/__init__.template
@@ -3,5 +3,5 @@ import pkg_resources
 
 try:
     __version__ = pkg_resources.get_distribution(__name__).version
-except:
+except pkg_resources.RequirementParseError:
     __version__ = 'unknown'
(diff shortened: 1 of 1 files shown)


FlorianWilhelm:
Hi @smartsammler, you are totally right about the fact that in almost all cases it is better to specify the actual exception one would like to catch in order to avoid catching unintended exceptions which leads to hard to find errors. Explicit is better than implicit.

Having said that, I think that this is actual one of the rare exceptions to the rule above. The `__init__.py` in the root of a Python package is like a backbone. If something breaks at that point, your Python package becomes almost useless. Since `pkg_resources.get_distribution` could throw more exceptions than just the `RequirementParseError`, it's best to have a valid fallback in all cases.

smartsammler:
I see your point and I think it is better to catch all exceptions then. But to provide a __init__.py file that does not need to be edited by hand if one uses the `--with-tox` option which will make flake8 complain about error E722, I added `# noqa`. Is this ok or should one only add this if one uses the `--with-tox` flag?

FlorianWilhelm:
Hi @smartsammler, this is fine thanks.

PR #137, opened 2017-12-14
Update Python versions in tox.ini template

smartsammler:
Python 3.3 is outdated since 2017-09-29, but Python 3.6.3 is nearly one year old.


FlorianWilhelm:
Thanks a lot!

smartsammler:
You're welcome!

Issue #155, opened 2018-04-09
Pyscaffold and Setuptools < 30.3.0


jleni:
When installing pyscaffold I get the following:

      File "/tmp/easy_install-c56y92hx/PyScaffold-3.0.2/setup.py", line 76, in bootstrap_cfg
      File "/tmp/easy_install-c56y92hx/PyScaffold-3.0.2/src/pyscaffold/utils.py", line 280, in check_setuptools_version
    pyscaffold.exceptions.OldSetuptools: Your setuptools version is too old (<30.3.0). Use `pip install -U setuptools` to upgrade.

Should setuptools >=30.3.0 should be included in the setup.cfg of the package?

FlorianWilhelm:
Thank you. Not sure if there was a reason I didn't specify this dependency.
Let me check when I am back from vacation. Thanks for your PR!

FlorianWilhelm:
Well, it definitely makes sense. It's merged. Thank you!

jleni:
Is there any chance we could have pyscaffold 3.0.3 released with the fix?

PR #156, opened 2018-04-09
Adding setuptools dependency

jleni:
This PR adds setuptools>=30.3.0 as a dependency during installation to solve issue #155


FlorianWilhelm:
Thanks!

PR #157, opened 2018-04-11
WIP: Fix build_sphinx command for sphinx>1.5

### LINKED ISSUE #152: Backport Sphinx 1.6 support to PyScaffold 2.5
Right now PyScaffold 2.5.x can only handle Sphinx up to version 1.5 since there was an API change. PyScaffold 3 was changed accordingly and this change could be also back-ported to PyScaffold 2.5.x.

cpaulik:
The new apidoc command does not take the executable as a first argument.

It also takes the excludes as a positional argument after the module directory.

Fixes #152


FlorianWilhelm:
@cpaulik, the problem was fixed with an update of pbr. Thanks for pointing that out @cpaulik!

Issue #158, opened 2018-04-14
clean up imports in __init__.py


reuvenpo:
Currently in the `__init__.py ` file there are two names imported from `setuptools`. These names should probably be `del`ed before the module finishes running to clean up the package's namespace.

For example:

```python
# -*- coding: utf-8 -*-
from pkg_resources import get_distribution, DistributionNotFound

try:
    # Change here if project is renamed and does not equal the package name
    dist_name = 'test-pscf-proj'
    __version__ = get_distribution(dist_name).version
except DistributionNotFound:
    __version__ = 'unknown'
finally:
    del get_distribution, DistributionNotFound
```

This is because if anyone runs `from myproj import *`, they're going to get `get_distribution` and `DistributionNotFound` into their module namespace as well.

By the way, looking at this file, the `dist_name` can probably be replaced with `__name__` which removes the need to explicitly spell it out in the `__init__` file.

Also, is `'unknown'` an appropriate value here? it's not a valid PEP440 version string if i understand correctly. Maybe `None` is more appropriate? I'm not sure what's the right thing to do here but `'unknown'` smells wrong to me.

FlorianWilhelm:
Hi @reuvenpo, I got the point with `finally: del get_distribution, DistributionNotFound`. Thanks. Would you make a small PR for that?

Regarding the usage of `__name__`. This is how it was done before and it doesn't work if your project has a different name compared to the package you are shipping. Common examples are for instance scikit-learn that ships a package sklearn and pytorch that ships a package torch. 
In this case `__name__` would be resolve into the package and `get_distribution` doesn't find the distribution. 

Regarding the `unknown`. I like to have the same data type here, namely string, so that the user doesn't always need to check for `None` when `__version__` is used. Otherwise this could lead to really strange bugs if for some reason (maybe a bug in setuptools or setuptools_scm) the right version is not resolved and the user's code breaks because it doesn't handle `None`. Does this make sense? Do you have a better PEP440 string in mind?

reuvenpo:
I'll make a pull request in a bit.

I got your point on the `__name__`. Didn't really consider that use-case.

As for the version name, if setuptools or setuptools_scm had a bug in it, wouldn't that mean we wouldn't get to this line executing, or alternatively we'll just get a malformed version string, which would silently be assigned to `__version__`.

[...]

PR #159, opened 2018-04-14
Added a namespace cleanup to the __init__.py template

reuvenpo:
This is a merge request related to issue #158.
previously, the in __init__.py template, several names from setuptools
remain in the namespace of the package, which later leakes to whomever
imports the package. Now these names are deleted from the module's
globals before the module finishes loading.


reuvenpo:
i think this change isn't significant enough to add myself to the contributors list :)

PR #165, opened 2018-06-01
Fix pre-commit-config and pre-commit extension templates

abravalheri:
Due to changes in the new versions released for the `pre-commit` package since the `pre-commit-config.yaml` file was written, some of the hooks were broken.

This PR is an attempt to have working and better hooks for PyScaffold. Additionally, similar configurations are introduced in the templates for the `pre-commit` extension.

A new .isort.cfg template is also introduced to facilitate using the `isort` hook included by default in the pre-commit configuration template.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.py[cod]
 *.so
 *.cfg
+!.isort.cfg
 !setup.cfg
 *.orig
 *.log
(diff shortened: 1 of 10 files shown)


FlorianWilhelm:
Wow, thanks a lot for this nicely documented code.

Issue #119, opened 2017-09-25
`test_install.py` not isolated properly


abravalheri:
The test functions in this file install packages directly in the environment they run, therefore "sharing" state among themselves.

This makes this test not thread-safe and prevents the usage of pytest-xdist to speed up the local/manually triggered verification and CI.

(Maybe  [pytest-virtualenv](https://github.com/manahl/pytest-plugins/tree/master/pytest-virtualenv) can be used to create a separated env for each test?)

abravalheri:
By the way, another thing preventing the usage of `pytest-xdist`, is that it doesn't support disabling output capture, according to [this issue](https://github.com/pytest-dev/pytest/issues/1991) (and I am using it to test the `termui`).

abravalheri:
#168 fix this issue, everything now is isolated under a venv :wink:

PR #166, opened 2018-06-01
Remove capsys.disabled

abravalheri:
One option to speed up tests (nowadays PyScaffold test suite is very slow) is to run them in parallel, however there are a few things preventing us to do that (see #119).

This PR tries to solve the [`capsys.disabled` limitation](https://github.com/pytest-dev/pytest/issues/1991), by not using it.

### DIFF
--- a/tests/test_termui.py
+++ b/tests/test_termui.py
@@ -2,7 +2,6 @@
 # -*- coding: utf-8 -*-
 from __future__ import absolute_import
 
-import sys
 from imp import reload
 
 from six import StringIO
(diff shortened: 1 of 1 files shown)


FlorianWilhelm:
@abravalheri Thanks! Just for my understanding, is this limitation of `capsys.disabled` something that will be fixed in the future? Should we maybe add a comment to `fake_tty` that this is actually a workaround for the limitations of `capsys.disabled`?

abravalheri:
@FlorianWilhelm , according to [pytest#1991](https://github.com/pytest-dev/pytest/issues/1991), it seems that they still don't see a way of enabling `capsys.disabled` and `xdist` to work togheter.

So I think the best is to add the comment as you have suggested. I can provide another PR :)

PR #168, opened 2018-06-04
Make `test_install` isolated by running everything inside separated virtualenvs

abravalheri:
This PR tries to solve #119 as effort to make the tests able to run in parallel.
(Althought I don't know if the extra trouble of creating multiple venvs is worth the gain in speed we can have).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ script:
   - tox
   - ./tests/system_test.sh
 after_success:
-  - if [[ "$COVERAGE" == "true" ]]; then py.test tests; coveralls || echo "failed"; fi
+  - if [[ "$COVERAGE" == "true" ]]; then tox; coveralls || echo "failed"; fi
 cache:
   apt: true
   pip: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ script:
   - tox
   - ./tests/system_test.sh
 after_success:
-  - if [[ "$COVERAGE" == "true" ]]; then tox; coveralls || echo "failed"; fi
+  - if [[ "$COVERAGE" == "true" ]]; then coveralls || echo "failed"; fi
 cache:
   apt: true
   pip: true
--- a/setup.cfg
+++ b/setup.cfg
@@ -33,7 +33,7 @@ include_package_data = True
 package_dir =
     =src
 install_requires = setuptools>=30.3.0
-tests_require = pytest; pytest-cov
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -46,7 +46,7 @@ script:
   - tox
   - ./tests/system_test.sh
 after_success:
-  - if [[ "$COVERAGE" == "true" ]]; then py.test tests; coveralls || echo "failed"; fi
+  - if [[ "$COVERAGE" == "true" ]]; then tox; coveralls || echo "failed"; fi
```
abravalheri: Sorry @FlorianWilhelm, I might have asked this in the past, but what would be the implications of removing this second round of executing the tests and just running `coveralls`?

This help to speed up the tests to run in the CI...

FlorianWilhelm: Hm.. actually I don't know anymore, maybe it's a copy & paste mistake. Does it also just work by only calling `coveralls`? If this is the case we should remove it.

setup.cfg
```diff
@@ -46,7 +46,7 @@ all = django; cookiecutter
 
 [test]
 # py.test options when running `python setup.py test`
-addopts = tests --verbose
```
FlorianWilhelm: Are you sure about removing `tests` at that point? Don't we need to pass the folder to the pytest-runner?

tox.ini
```diff
@@ -5,9 +5,10 @@ minversion = 1.8
 passenv =
     HOME
 commands =
-    py.test tests {posargs}
```
FlorianWilhelm: Also here, don't we need to pass the folder so that py.test just doesn't start looking into each directory?

abravalheri: According to the [docs](https://docs.pytest.org/en/latest/reference.html#confval-testpaths), if we set the `testpaths` options correctly, it is possible to restrict the places where pytest searches by default, and therefore we don't need to pass the folder directly to the command...

The advantage of this approach is that one can easily overwrite the files being tested without touching `tox.ini` and `setup.cfg`.

For example, I was doing `tox -- tests/test_install.py`, to have a quick feedback and avoid running everything all the time.

.travis.yml
abravalheri: Let's check :)

abravalheri: Well, seems to be working: https://coveralls.io/builds/17322542

tox.ini
FlorianWilhelm: Ah, nice. Didn't know that!

PR #169, opened 2018-06-04
Add a note about xdist/capsys.disabled limitation

abravalheri:
As discussed in #166 ...


FlorianWilhelm:
Thanks! You can also just push little changes like that directly on the master. I think you got the permission for it.

abravalheri:
Well, it think so, it's been a while since I had contributed, so I thought
it was better to ask than abuse :P

Thanks Florian!

--
Anderson

PR #171, opened 2018-06-05
Reorganize checks to make them more flexible

abravalheri:
As commented in #102, this PR tries to reorganize the checks prior to the materizalization of the scaffold structure.

They can be now splitted in 2 groups:
1. Pure option consistency checking
2. Checks that involve the underlying file system.

By postponing the second group, extensions have a chance to change the file system making it consistent before the check is done.

(In summary this PR doesn't change behavior, but add flexibility by reordering the actions)

### DIFF
--- a/src/pyscaffold/api/__init__.py
+++ b/src/pyscaffold/api/__init__.py
@@ -156,7 +156,6 @@ def get_default_options(struct, opts):
     # Initial parameters that need to be provided also during an update
     opts = DEFAULT_OPTIONS.copy()
     opts.update(given_opts)
-    project_name = opts['project']
     opts.setdefault('package', utils.make_valid_identifier(opts['project']))
     opts.setdefault('author', info.username())
     opts.setdefault('email', info.email())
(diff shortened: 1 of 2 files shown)


src/pyscaffold/api/__init__.py
```diff
@@ -206,10 +218,11 @@ def verify_options_consistency(struct, opts):
                 "Directory {dir} already exists! Use the `update` option to "
                 "update an existing project or the `force` option to "
                 "overwrite an existing directory.".format(dir=opts['project']))
-    if not utils.is_valid_identifier(opts['package']):
-        raise InvalidIdentifier(
-            "Package name {} is not a valid "
-            "identifier.".format(opts['package']))
+    else:
```
FlorianWilhelm: Wouldn't here an `elif` make sense to avoid another level of indentation?

FlorianWilhelm:
Thanks!

src/pyscaffold/api/__init__.py
abravalheri: Done!

Issue #170, opened 2018-06-05
Make tests "thread-safe" again!


abravalheri:
After the last changes from #166 and #168, we are halfway through being able to setup `pytest-xdist` (and I hope this is enough to speed up the test suite at least a little bit).

The tests preventing us to of doing that are: 
- `tests/extensions/test_cookiecutter.py : test_cli_with_cookiecutter`
- `tests/extensions/test_cookiecutter.py : test_create_project_with_cookiecutter`
- `tests/testutils.py : test_pretend_chdir`

It seems that these failures cannot be detected with a lower number of parallel workers executing.
I could manage to see then in the CI running 1 process per core, and locally running 16 processes despite of the number of cores.

abravalheri:
Turns out that there are 2 important and hidden pieces of shared state between the processes causing this trouble:

1. Cookiecutter uses shared "cache" directories (but it is configurable, so we can workaround).
2. In Python the logger object is global. Despite of being thread-safe (they have a locking mechanism), some tests I have designed check the log in a order-dependent way (still not quite sure about this one, since xdist uses processes, but some stream might be shared).

abravalheri:
The issue was addressed in the last few merge requests :)

PR #173, opened 2018-06-06
Enforce isolation for cookiecutter tests

abravalheri:
Create and configure custom, per-test, cookiecutter "cache/storage" dirs.

This PR is related to #170  and tries to make cookiecutter work nicely in a multi-process/thread test environment.


FlorianWilhelm:
Thanks

Issue #120, opened 2017-09-25
Suggestion for system tests


abravalheri:
Right now system tests are performed in a ad-hoc fashion with a simple bash script.

A common approach to add more organization / test capabilities is to use a bash test framework like [bats](https://github.com/sstephenson/bats).

Another approach would be also using pytest with the help of some plugins such as [pytest-virtualenv](https://github.com/manahl/pytest-plugins/tree/master/pytest-virtualenv) and [pytest-shutil](https://github.com/manahl/pytest-plugins/tree/master/pytest-shutil). I personally have never used these plugins,  but if they work, keeping everything in python would be nice.

FlorianWilhelm:
Sounds like a good idea, would [tox](http://tox.readthedocs.io/en/latest/index.html) not help regarding the virtualenv problem?

abravalheri:
tox would definitely help. But from what I read from the documentation the pytest-virtualenv package is even more useful since it enables calling executables from a python function as if they were installed with pip.

FlorianWilhelm:
Ah, I didn't know that, so far I have only used tox but of course we should opt for the most suitable but also reliable option.

abravalheri:
Solved in #185

PR #174, opened 2018-06-08
Isolate logging in tests

abravalheri:
This PR is an attempt to solve issue #170.

I have removed order-dependent test helpers and used unique names instead. Additionally, this PR introduces a new context fixture in an attempt to better isolate the logging usage in the tests.

An error (previously undetected) in the chdir logging is also corrected.

However, the results of the tests are still non-deterministic: in some occasions `caplog.text` is unexpectedly blank as indicated by the CI logs. Any tip/hunch/help on this issue is more then welcome, I couldn't completely figure it out yet.

(Who would say, logging would be the most difficult thing to test??)

Regarding performance: in the CI, the speed improvement is not very noticeable (in part because of the setup time and the system tests in bash that do not run in parallel - running everything inside pytest as suggested in #120 could help).

### DIFF
--- a/src/pyscaffold/utils.py
+++ b/src/pyscaffold/utils.py
@@ -13,15 +13,15 @@ import sys
 from contextlib import contextmanager
 from operator import itemgetter
 
+from .contrib.setuptools_scm.version import VERSION_CLASS
+from .contrib.six import PY2
 from .exceptions import InvalidIdentifier, OldSetuptools
 from .log import logger
 from .templates import licenses
-from .contrib.six import PY2
-from .contrib.setuptools_scm.version import VERSION_CLASS
 
 
 @contextmanager
-def _chdir_logginng_context(path, should_log):
+def _chdir_logging_context(path, should_log):
     """Private auxiliar function for logging inside chdir"""
     if should_log:
         logger.report('chdir', path)
@@ -51,7 +51,7 @@ def chdir(path, **kwargs):
     curr_dir = os.getcwd()
 
     try:
-        with _chdir_logginng_context(path, should_log):
+        with _chdir_logging_context(path, should_log):
             if not should_pretend:
                 os.chdir(path)
             yield
--- a/src/pyscaffold/utils.py
+++ b/src/pyscaffold/utils.py
@@ -21,6 +21,17 @@ from .contrib.setuptools_scm.version import VERSION_CLASS
 
 
 @contextmanager
+def _chdir_logginng_context(path, should_log):
+    """Private auxiliar function for logging inside chdir"""
+    if should_log:
+        logger.report('chdir', path)
+        with logger.indent():
+            yield
+    else:
+        yield
+
+
+@contextmanager
 def chdir(path, **kwargs):
     """Contextmanager to change into a directory
 
@@ -38,15 +49,11 @@ def chdir(path, **kwargs):
     #   (after all, this is the primary purpose of pretending)
 
     curr_dir = os.getcwd()
-    if not should_pretend:
-        os.chdir(path)
 
     try:
-        if should_log:
-            logger.report('chdir', path)
-            with logger.indent():
-                yield
-        else:
+        with _chdir_logginng_context(path, should_log):
+            if not should_pretend:
+                os.chdir(path)
             yield
     finally:
         os.chdir(curr_dir)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -13,9 +13,6 @@ from shutil import rmtree
 
 import pytest
 
-from pyscaffold.exceptions import ShellCommandException
-from pyscaffold.log import ReportFormatter
-
 from .helpers import uniqstr
 
 try:
@@ -43,6 +40,14 @@ def set_writable(func, path, exc_info):
         func(path)
     else:
         raise
+
+
+def command_exception(content):
+    # Be lazy to import modules, so coverage has time to setup all the
+    # required "probes"
+    # (see @FlorianWilhelm comments on #174)
+    from pyscaffold.exceptions import ShellCommandException
+    return ShellCommandException(content)
 
 
 @pytest.fixture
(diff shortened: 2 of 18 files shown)


src/pyscaffold/utils.py
```diff
@@ -20,6 +20,17 @@
 from .contrib.setuptools_scm.version import VERSION_CLASS
 
 
+@contextmanager
+def _chdir_logginng_context(path, should_log):
```
FlorianWilhelm: logginng? Is this a typo or intended?

tests/helpers.py
```diff
@@ -0,0 +1,25 @@
...
+
+from six.moves import xrange
+
+
+def randstr(n=5):
+    """Generates a random string with n ascii_letters"""
+    return ''.join(choice(ascii_letters) for _ in xrange(n))
+
+
+def uniqstr():
```
FlorianWilhelm: Aren't there any functions in Python's standard module `uuid` that could be used? Just wondering.

tests/conftest.py
```diff
@@ -59,6 +62,38 @@ def logger():
     return pyscaffold.log.logger
 
 
+@pytest.fixture
+def isolated_logger(logger):
```
FlorianWilhelm: For future contributors (and future me ;-) I think it makes sense to describe better what is done here conceptually. As far as I see it you solve the whole singleton logging issue by creating multiple ones with unique ids, right? Somewhere it should be described.

tests/helpers.py
abravalheri: UUID could be used, but some extra logic might be required, because I believe python logging required valid identifiers (something like stripping dashes and prepending letters).

If you prefer, I can go in that direction.

src/pyscaffold/utils.py
abravalheri: Typo :)

Thanks for the review!

abravalheri:
Hi Florian, thanks for the review! Please find my answers in line:

```
This seems to be the final step for complete isolation, right?
Can you yet tell what speedup in the unit-tests we get?
```
Final step \o/.  In the CI the speed up seems to be negligible (something around 1 minute, but I haven't done proper statistics). When running locally, the speed up is more noticeable when we use more than 10 processes, even in machines with a lower number of cores (I am testing in a VM with 2).

I believe this is because we do a lot of IO. However changing dynamically the number of process in xdist is not easy (I would like to have something like `max(10, numcores)` or `numcores + 5` to fully take advantage of the CI) - right now the configuration is to use the number of cores available.

Pytest allow us [to do so](https://docs.pytest.org/en/latest/example/simple.html#dynamically-adding-command-line-options), but not from `conftest`. (I am still investigating if we need an setuptools entrypoint for that or not - if we need I don't think it is worthy the trouble to create a pytest plugin).

```
And has it any more downsides like unreadable logs?
I could imagine that if several unit tests run in parallel you get an confusing output.
```
Writing tests become more difficult, but not that much. (Just when you are using shred resources, like files - I think I fixed the logs now!).
It turns out pytest do a good job organizing the test logs... You can check out the CI ones to get an idea.

```
Can it easily be deactivated if there is any need?
Would be nice to also that that somewhere, how to do it.
```
It would require a custom command line parameter, something similar to the pytest configuration/plugin I have previously mentioned in my first comment. It can be quite difficult to do.

Last comments:
- I might have fixed the non-deterministic behaviour :)
- If we decide to not do parallel testing, I still think some changes in this PR are welcome, and having isolation is something good.

tests/helpers.py
FlorianWilhelm: No, I was just wondering if it was considered. I always tend to prefer simpler solution. So if using `uuid` would result in something even simpler but I guess we should keep it as it is right now.

FlorianWilhelm:
Thanks @abravalheri, isolation is definitely something we want and the same goes for deterministic behaviour. 

I took a look at the Travis logging and something I noticed is a lot of messages like:
```
tests/api/test_helpers.py::test_merge_rules_just_in_merged Coverage.py warning: Module pyscaffold was previously imported, but not measured (module-not-measured)
Coverage.py warning: Module pyscaffold was previously imported, but not measured (module-not-measured)
Coverage.py warning: Module pyscaffold was previously imported, but not measured (module-not-measured)
```
as well as:
```
Module already imported so cannot be rewritten: coverage
```
Is this related to xdist?

tests/helpers.py
abravalheri: I did some experiments, and it seems that, despite [suggesting it](https://docs.python.org/3.6/library/logging.html#logger-objects), the `module identifier` format is just recommended, not required.

I will try to remove `uniqstr` in favor of `uuid4` and see what happens :)

Thanks for the suggestion.

abravalheri:
Hi @FlorianWilhelm,

The warning `Module already imported so can not be re-written` at the end of the tests, is something that appears to be quite [common](https://github.com/pytest-dev/pytest-cov/issues/148) when using xdist, and is related to the random order pytest imports its plugins (the solution is something still being [debated](https://github.com/pytest-dev/pytest/issues/935)).

My pragmatic fix for the warning was removing the `-p coverage` from the list of arguments in `setup.cfg`.

The warning `Module pyscaffold was previously imported, but not measured`, seems to be related to importing `pyscaffold` in `conftest.py`. I reckon pytest loads conftest first because of the plugins and configurations that can be there, but then xdist have to load the entire test environment again for each process, setting up the "coverage probes". At this point, `pyscaffold` is already cached in memory.

My pragmatic fix for that is lazy loading everything inside conftest.

abravalheri:
@FlorianWilhelm, since it is not easily to "opt-out" of xdist if we include its configuration on `setup.cfg`, I have changed to a `opt-in` approach and added some instructions to the `CONTRIBUTING.md`.

I hope this is a good compromise.

Soon I will provide you some numbers about the real difference in time of using xdist or not.

FlorianWilhelm:
@abravalheri, looks cool! Thanks for your work!

Issue #161, opened 2018-05-16
tox.ini template is slow at running tests


aspiers:
IIUC [`src/pyscaffold/templates/tox_ini.template`](https://github.com/blue-yonder/pyscaffold/blob/master/src/pyscaffold/templates/tox_ini.template) causes tox to do an sdist install even though the tests are unlikely to need it.  This can slow testing down a lot.  If I'm correct, this should be fixed by adding `usedevelop = True` to the `[testenv]` section:

http://tox.readthedocs.io/en/latest/example/general.html#avoiding-expensive-sdist

See also #160.

FlorianWilhelm:
Thanks @aspiers for your suggestion. Surely in large projects the `usedevelop = True` option could help to speed up things but for smaller projects the overhead is quite minimal anyway. The downside to it is that the actual packaging and installing is no longer tested by tox. 

So imagine you have forgotten to add some data file to the repository and thus it won't be in the source distribution. In this case with `usedevelop = True` everything should run smoothely but without it you will see the error that the data file is missing as it should be.

What do you think about a compromise here. We could add the commented line `# usedevelop = True` to the template together with a comment to uncomment it in order to speed up the unit tests if necessary.

FlorianWilhelm:
@abravalheri, I added it to the template as a comment as discussed and made a PR. I think we can close this now.

aspiers:
Sorry for the slow reply.  I didn't personally find that for smaller projects the overhead was minimal - in fact it was a small project of mine which prompted submitting this ticket.  However I think the compromise here is good, so thanks a lot!

PR #178, opened 2018-06-13
Tox templates

### LINKED ISSUE #160: tox.ini template is needlessly slow at linting with flake8
IIUC [`src/pyscaffold/templates/tox_ini.template`](https://github.com/blue-yonder/pyscaffold/blob/master/src/pyscaffold/templates/tox_ini.template) causes `tox -e flake8` to do an install even though flake8 is linting the source files directly.  If I'm correct, this should be fixed by adding `skip_install = True` to the `[testenv:flake8]` section.

Also note that this should work even if `[testenv]` has `usedevelop = True`: https://github.com/tox-dev/tox/pull/572

See also #161.

FlorianWilhelm:
This fixes #160 and #161. 

@abravalheri Is this okay?

### DIFF
--- a/src/pyscaffold/templates/tox_ini.template
+++ b/src/pyscaffold/templates/tox_ini.template
@@ -8,6 +8,8 @@ envlist = py27,py34,py35,py36,flake8
 skip_missing_interpreters = True
 
 [testenv]
+# uncomment `usedevelop` for faster tests without package builds & installs
+# usedevelop = True
 changedir = tests
 commands =
     py.test {posargs}
@@ -17,6 +19,7 @@ deps =
     -r{toxinidir}/requirements.txt
 
 [testenv:flake8]
+skip_install = true
 changedir = {toxinidir}
 deps = flake8
 commands = flake8 setup.py ${package} tests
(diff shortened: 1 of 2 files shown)


src/pyscaffold/templates/tox_ini.template
```diff
@@ -8,6 +8,8 @@ envlist = py27,py34,py35,py36,flake8
 skip_missing_interpreters = True
 
 [testenv]
+# uncomment `usedevelop` for faster tests without package builds & installs
```
abravalheri: How about adding:
```
...
# but please be aware that you will not be testing the package as if it was installed by the final user
# see http://tox.readthedocs.io/en/latest/example/general.html#avoiding-expensive-sdist
```

Or something like that?

FlorianWilhelm: I would like to avoid two line comments and too many deep links into the documentation of others that get outdated quite easily. The main tox documentation is already in the head of the file addressed. What do you think about the following change below?

abravalheri: :+1: 
Very clear, it should be enough!
Thanks

PR #179, opened 2018-06-13
Add pre-commit to Travis job

abravalheri:
As discussed in #160, instead of calling flake8 directly or from tox, if we call `pre-commit run --all-files` we take advantage of all the other checks in pre-commit (see `.pre-commit-config.yaml` for a list), which is even nicer.

Of course, a little round of minor changes in the current files to adequate them to the conventions coded in the config files (such as `.isort.cfg`) was necessary, but the good part is that, with the result, if the contributor follows all the steps in the `CONTRIBUTING.rst` file, no surprises will appear (such as a failing pre-commit hook), and the coding conventions will be always consistent.

I also updated the format of the `.pre-commit-config.yaml`  to the [latest version](https://pre-commit.com/#pre-commit-configyaml---top-level), and configured a list of paths that are not part of the project itself to be ignored in all the checks.


FlorianWilhelm:
@abravalheri, cool! Thank you.

PR #184, opened 2018-06-18
Increase minimum setuptools version

FlorianWilhelm:
This should be necessary in order to allow the inclusion of ``entry_points`` in ``setup.cfg``. Additionally, some already resolved issue were added to the ``CHANGELOG.rst``.

@abravalheri It seems that ``options.entry_points`` works already with setuptools version 31 and above. At least it does on my machine. Could you check if it is the same for you?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,9 @@ Development version
 Version 3.1
 -----------
 
+- Moved ``entry_points`` and ``setup_requires`` to ``setup.cfg``, issue #176
+- Updated ``travis.yml`` template, issue #181
+- Set ``install_requires`` to setuptools>=31
 - Better isolation of unit tests, issue #119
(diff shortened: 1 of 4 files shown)


abravalheri:
Hi @FlorianWilhelm , I have tested it using the following script:

```bash
#!/user/bin/env bash
set -ex

cd /tmp
conda config --add channels conda-forge
conda create -p /tmp/.venv --yes python=3.4 pip "setuptools>=31.0,<32"
source activate /tmp/.venv
git clone https://github.com/blue-yonder/pyscaffold.git
cd pyscaffold
python setup.py develop
cd /tmp
putup myproj
cd myproj
sed -i.bkp 's/\(\[options.entry_points\]\)/\1\nconsole_scripts=\n    fibonacci = myproj.skeleton:run/' setup.cfg
python setup.py install
which fibonacci
```

End everything seems to be fine.

PR #186, opened 2018-07-01
pre-commit: No need to exclude untracked things

asottile:
pre-commit will only ever call linters with files that are checked in


FlorianWilhelm:
@abravalheri Thanks!

PR #187, opened 2018-07-05
Remove never running tests

abravalheri:
This PR documents the Cookiecutter/update error, removes pointless tests and add useful ones, as discussed in #185 comments.


FlorianWilhelm:
Nice! Thanks a lot!

Issue #140, opened 2018-01-15
Pipenv Support


sbeleidy:
Are there any plans to support [pipenv](https://docs.pipenv.org)?

FlorianWilhelm:
Hi, so far I haven't used it. Supporting it would mean to provide a basic `Pipfile`, right? Should be no problem with the new extension system. A `--pipenv` flag would then create a basic `Pipfile` instead of `requirements.txt`. 
Do you want to provide a PR?

sbeleidy:
I think it's a little more than that. Adding a `Pipfile` is a simple thing - the issue I'm not sure how to deal with is loading the dependencies from the `Pipfile` to be used in `setup.py`. When you install the package it should install all its dependencies as well.

Happy to provide a PR if you can direct me to the best way to deal with this.

FlorianWilhelm:
@sbeleidy, and don't think that setup.py should read the requirements from `Pipefile`. The requirements in the Python package defined by `setup.cfg` are different than the requirements needed to set up and run the project itself. Therefore I think an extension should just write a `Pipfile` with some sane defaults instead of the current `requirements.txt`.

sbeleidy:
@FlorianWilhelm - I think I'm a little confused. If your package has dependencies, you are setting those in setup.cfg? Do you have an example of that? 

I thought it would read the requirements.txt and add those packages.

FlorianWilhelm:
Sure, this behaviour was changed in version 3.0 of PyScaffold. As you can see in this [configuration example](https://pyscaffold.readthedocs.io/en/latest/configuration.html) there is a key `install_requires` where you would list the unpinned, direct requirements of your package. 

The `requirements.txt` itself it meant to help you set up your development or runtime environment pretty fast and there you would rather use pinned versions of your requirements and maybe even add the requirements of your requirements and so on to make sure the environment is reproducible. You would thereby use tools like [pip-tools](https://github.com/jazzband/pip-tools/) to manage your `requirements.txt`. The new `Pipfile` format replaces the old-school `requirements.txt` and is managed by `pipenv`.

HolgerPeters:
> I think it's a little more than that. Adding a Pipfile is a simple thing - the issue I'm not sure how to deal with is loading the dependencies from the Pipfile to be used in setup.py. When you install the package it should install all its dependencies as well.

[...]

PR #188, opened 2018-07-06
Add docs about how to use external dependency managers (e.g. Pipenv) with PyScaffold

abravalheri:
This PR tries to implement the proposal discussed in #140


FlorianWilhelm:
@abravalheri, thanks! The new dependency page reads really well!

Issue #177, opened 2018-06-13
Drop Python 2 support in PyScaffold v3.1


FlorianWilhelm:
PyScaffold is a tool for creating new Python projects and Python 3 surely has been for quite a while the right version to start a new project. Also due the a [bug](https://github.com/pypa/setuptools/issues/1136) in `setuptools` the current version 3.0 of PyScaffold has no Python 2 support right now and there were no complaints about this. 
We could officially drop the Python 2 support in PyScaffold 3.1 and get rid of `six`, unnecessary `__future__` imports in templates and the code of PyScaffold itself. This would clean things up.

abravalheri:
To be sincere, since the test suite is intrinsically broken with 2.7, all the last changes might be incompatible  anyway...

FlorianWilhelm:
Okay, sounds like we have a decision then ;-)

FlorianWilhelm:
Since we have never had Python 2 support in version 3.0 we can drop it in version 3.1 officially without violating semantic version. I wouldn't want to wait for version 4.0 for that.

FlorianWilhelm:
Since the `six` and all `__future__` imports where removed this ticket can be closed. As we go we should start using more Python 3 features and make things even more pythonic.

abravalheri:
Unfortunately we have to wait 3.7 to be the base line to use f-strings :stuck_out_tongue_closed_eyes:

PR #189, opened 2018-07-09
Drop PY2 support

FlorianWilhelm:
This should partially address issue #177 by removing `six` and `__future__` imports. The more complicated part would then be to check the source code step by step if newer Python 3 language features can be used to make things even more pythonic.


FlorianWilhelm:
@abravalheri, good question actually. I think it should be possible with PyScaffold 3 using universal wheels to built a package that runs under Python 2 with the help of `six`. We should check it before we add a note like that into the docs. It's definitely a good point. It's actually a good FAQ!

abravalheri:
Should we add a system test to ensure that and make it an official feature?

FlorianWilhelm:
Yes, a system test would be the best solution I guess although complicated because one would create with PY3 and run with PY2. So a change of the interpreter during the test? If it is too much hassle I think it's best to test it just once.

abravalheri:
Maybe it is not exactly straightforward doing that (I just opened a `pytest-virtualenv` issue to check if that is possible). I think it would require the use of bash scripts again.

Issue #139, opened 2017-12-28
Rework the setup.cfg with the help of configparser


FlorianWilhelm:
Using `string.template` for more complicated templates like `setup.cfg` is quite cumbersome and would also not allow to update `setup.cfg` in case of new versions. We should use `configparser` instead to create `setup.cfg` as well as to update it later on.

This can be postponed to version 3.1 since an update from pre 3.0 versions will not be possible anyway.

FlorianWilhelm:
This issue seems to be actually way more complicated than initially thought. After some investigation there seems to be no Python package right now that allows to read, manipulate and write a config file without potentially changing the order of the key/value pairs and dropping all comments of the original config file.

I appreciate any hints on this.

abravalheri:
The order of the key/value pairs seems to be easy to be kept if an OrderedDict is used, however keeping the comments is a little bit trickier...

In order to write comments, PyScaffold would have to use some [hacks](https://stackoverflow.com/questions/6620637/writing-comments-to-files-with-configparser), but storing the comments in a read operation, changing the options and then writing seems pretty much impossible.

In order to do that, PyScaffold would need something that reads the `cfg` file and returns an AST, but I don't think this is somehow easily available.

Are there other alternatives than writing a parser by hand?

FlorianWilhelm:
Maybe the alternative is to check if large parts of Python's internal `configparser` could be used. Using an `OrderedDict` at least keeps the order when reading a config file. No we would just need to include comments somehow and make sure that they also are written back.

FlorianWilhelm:
I am trying to implement the needed functionality in an own project named [ConfigUpdate](https://github.com/pyscaffold/configupdater) that can later be included as `contrib` package. It's making good progress but needs some more time.

FlorianWilhelm:
@abravalheri, I just released a first `0.1` version of [ConfigUpdater](https://github.com/pyscaffold/configupdater). Let me know what you think. 

My next step would be now to include it into PyScaffold and use it to allow updating `setup.cfg` from version 3.0 to version 3.1, i.e. moving `setup_requires` and `options.entry_points` over as well as updating the `pyscaffold` section.

FlorianWilhelm:
Done

PR #194, opened 2018-07-28
Allow updating of update.cfg

FlorianWilhelm:
This PR is a proposal how to do the updating from one PyScaffold version to another with a focus on `setup.cfg`. It thus also addresses #139 since it uses [configupdater](https://github.com/pyscaffold/configupdater) to accomplish all this. I haven't yet created any additional unit tests to avoid any unnecessary work if this is a dead end but so far my local tests look quite good and it does what it is intended to do. 
@abravalheri What do you think about this approach?

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,7 +1,7 @@
 # .coveragerc to control coverage.py
 [run]
 branch = True
-source = */pyscaffold/*
+source = pyscaffold
 omit =
     */PyScaffold*.egg/*
     */pyscaffold/contrib/*
--- a/src/pyscaffold/update.py
+++ b/src/pyscaffold/update.py
@@ -8,7 +8,7 @@ from functools import reduce
 from os.path import exists as path_exists
 from os.path import join as join_path
 
-from packaging.version import parse as parse_version
+from pkg_resources import parse_version
 
 from . import __version__ as pyscaffold_version
 from .api import helpers
--- a/src/pyscaffold/update.py
+++ b/src/pyscaffold/update.py
@@ -4,10 +4,11 @@ Functionality to update one PyScaffold version to another
 """
 import os
 from configparser import ConfigParser
-from distutils.version import LooseVersion
 from functools import reduce
 from os.path import exists as path_exists
 from os.path import join as join_path
+
+from packaging.version import parse as parse_version
 
 from . import __version__ as pyscaffold_version
 from .api import helpers
@@ -131,11 +132,11 @@ def get_curr_version(project_path):
         project_path: path to project
 
     Returns:
-        LooseVersion: version specifier
+        Version: version specifier
     """
     setupcfg = ConfigParser()
     setupcfg.read(join_path(project_path, 'setup.cfg'))
-    return LooseVersion(setupcfg['pyscaffold']['version'])
+    return parse_version(setupcfg['pyscaffold']['version'])
 
 
 def version_migration(struct, opts):
@@ -158,8 +159,8 @@ def version_migration(struct, opts):
 
     # specify how to migrate from one version to another as ordered list
     migration_plans = [
-        (LooseVersion('3.1'), [add_entrypoints,
-                               add_setup_requires])
+        (parse_version('3.1'), [add_entrypoints,
+                                add_setup_requires])
     ]
     for plan_version, plan_actions in migration_plans:
         if curr_version < plan_version:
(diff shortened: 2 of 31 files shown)


src/pyscaffold/update.py
```diff
@@ -0,0 +1,272 @@
...
+    Args:
+        struct (dict): previous directory structure (ignored)
+        opts (dict): options of the project
+
+    Returns:
+        tuple(dict, dict):
+            structure as dictionary of dictionaries and input options
+    """
+    setupcfg = read_setupcfg(opts['project'])
+    section_str = """[options.entry_points]
```
abravalheri: Maybe we should add this to a template file and also make the setup.cfg template to use this one?

abravalheri:
Hi @FlorianWilhelm, thank you very much for the implementation!
Regarding the test suite is always good to have one, but the fact that it doesn't break any existing test is already a good signal.

I haven't tested it locally, just looked at the diff and everything seems to be OK. If you want me to do a more complete review and also perform some manual tests. I will need some extra time :P

In my point of view now the only thing missing to consider the update feature more or less complete, is to solve the existing issue with Cookiecutter: it should never run on an update (so `cookiecutter`, and any other extension that behaves similarly - I guess `django` also - should never be included under `extensions` in `setup.cfg`).

src/pyscaffold/update.py
```diff
@@ -0,0 +1,272 @@
...
+    """
+    setupcfg = read_setupcfg(opts['project'])
+    comment = ("# DON'T CHANGE THE FOLLOWING LINE! "
+               "IT WILL BE UPDATED BY PYSCAFFOLD!")
+    options = setupcfg['options']
+    if 'setup_requires' in options:
+        return struct, opts
+
+    version_str = get_setup_requires_version()
+    (options['package_dir'].add_after
```
abravalheri: This fluent interface is very nice :+1:

src/pyscaffold/update.py
FlorianWilhelm: Yes, good point, will do that!

src/pyscaffold/update.py
FlorianWilhelm: Thanks!

FlorianWilhelm:
Thanks for your comments. Regarding `cookiecutter` and `django`. Couldn't we just let the extensions themselves check for `opts['update'] is True` and in that case just do nothing? In this case the extension itself is responsible what to do during an update. 

Regarding the version 3.1 release. The last issue we have for besides this one is the one that allows you to do an update with `putup -U ./`. I don't think this is so difficult to implement but will take some time. Do you think we should include it in 3.1 or postpone to 3.2?

src/pyscaffold/utils.py
```diff
@@ -295,3 +298,38 @@ def dasherize(word):
         input word with underscores replaced by dashes
     """
     return word.replace('_', '-')
+
+
+def get_id(function):
```
abravalheri: Well, since we are on python3-only now, we could just use `__qualname__`. It would be longer, but do the job nicely.

FlorianWilhelm: Good point, I removed `get_id`.

FlorianWilhelm: Hm.. just realised that `__qualname__` does not include the module name. So I kept `get_id`.

abravalheri: Sorry for the misleading comment :(
I was under the impression that it would include.

.coveragerc
```diff
@@ -1,7 +1,7 @@
 # .coveragerc to control coverage.py
 [run]
 branch = True
-source = */pyscaffold/*
```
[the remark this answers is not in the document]
FlorianWilhelm: Hi @abravalheri, not really. I what I did basically is that in case of `COVERAGE=1`, the unit tests are run again with pytest only to show the coverage correctly. I read somewhere on Stackoverflow that it is possible to map different paths to one and to completely circumvent this problem by some nifty `.coveragerc` settings. It has something to do with [paths section](https://coverage.readthedocs.io/en/coverage-4.5.1a/config.html?highlight=path#paths). Wanna look into it?

FlorianWilhelm:
@abravalheri, I am really sorry that I pointed you to something you have already tried and mentioned several times... I realised just lately what a pain tox and pytest-cov can be. 

In this PR I solved the problem by moving the `cwd` to the actual source code repository before running `coverage`. This way coverage creates those `.coverge*` result files in the right directory and they will be later collected and combined. Not sure that helps...

Thanks for providing a PR!

FlorianWilhelm:
If things get too complicated with that workaround we could also call `py.test` without `tox` in the end like it is done right now, of course with some additional runtime.

Issue #197, opened 2018-08-13
test_update.py failing


abravalheri:
When I try to run the test suite, but the repository is dirty, or I am some commits ahead of the commit I did `python setup.py egg_info --egg-base .` , I face an assertion error on the line 101 of `test_update.py`
```python
# ...
assert __version__ == str(self.pyscaffold_version())
# ...
```

Usually, the versions diff by a `.dirty` or `.dev<N>... != .dev<N+1>...` suffix .
The way I made it work was by running `python setup.py egg_info --egg-base .` again.
However always running this workaround can be a little bit annoying.

We could include this command in the `tox.ini`, but is it the best idea?
@FlorianWilhelm, what do you think?

FlorianWilhelm:
Thanks for the PR! I'll look into the `test_update` issue with a dirty workspace. I think we can just use the Version object and there is some kind of method that only returns (major, minor, patch). Can I just revert to the original `tox.ini` after that?

FlorianWilhelm:
BTW, if you pull from master, make sure you delete all egg files especially the one under src before you do `python setup.py egg_info --egg-base .` one time.

abravalheri:
Thanks @FlorianWilhelm! Should we also discard the dev notes on `CONTRIBUTING.rst`? Or instead keep the notes there and also add your comment about deleting the egg files (it might me useful)?

abravalheri:
After the last changes, the case when the repository is dirty was solved.
However, if the development version is commits ahead, we will also have problems.

After making a commit (changes in `tox.ini`), and running the test again, I had:

```
>       assert self.running_version.public == self.pyscaffold_version().public
E       AssertionError: assert '3.1b2.post0.dev1' == '3.1b2.post0.dev2'
E         - 3.1b2.post0.dev1
E         ?                ^
E         + 3.1b2.post0.dev2
E         ?                ^

tests/test_update.py:102: AssertionError
```

FlorianWilhelm:
Why should we delete the dev notes? I think it is better to extend them since it helps new developers to become a contributor. 
The  `self.running_version.public == self.pyscaffold_version().public` is expected. One could use `.base_version` which would result in `3.1` but then it has almost no more information. At that point I only wanted to check that the `pytest-virtualenv` plugin does not install pyscaffold with the help of PyPI which happened a lot at first. Maybe we should just drop the assert? You could comment it out with a note if you want.

abravalheri:
I agree in expanding the notes :)
[...]

PR #199, opened 2018-08-14
Polish solution presented for #197

abravalheri:
This PR relates to the on-going discussion on #197 

- Removes undefined testenv in tox.ini
- Includes a troubleshooting section in `CONTRIBUTING.rst`
- Change the comparison in `test_update` from `==` to `<=`
- Adds an assert that makes sure PyScaffold was installed from source in a editable mode, instead of PyPI


FlorianWilhelm:
Thanks!

PR #200, opened 2018-08-14
Implement workaround for wrong coverage reporting

abravalheri:
As discussed in:
https://github.com/blue-yonder/pyscaffold/pull/194#discussion_r209830873

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -93,22 +93,18 @@ setuptools_scm.local_scheme =
 
 [test]
 # py.test options when running `python setup.py test`
-addopts = --verbose
+addopts = --verbose --cov-report=term-missing
 extras = True
 
 [tool:pytest]
 # Options for py.test:
 # Specify command line options as you would do when invoking py.test directly.
-# e.g. --junitxml junit.xml in order to write a coverage file that can be read
-# by Jenkins.
+# e.g. --cov-report html (or xml) for html/xml output or --junitxml junit.xml
+# in order to write a coverage file that can be read by Jenkins.
 addopts =
     --cov pyscaffold
     --cov-config .coveragerc
-    --cov-report=
     --verbose
-# Please call `coverage combine .coverage` explicitly after running the tests.
-# You can display/generate a report with `coverage report -m` or `coverage xml`
-#
 # In order to use xdist, please add something like
 # --dist=load --numprocesses=16
 norecursedirs =
--- a/setup.cfg
+++ b/setup.cfg
@@ -99,14 +99,18 @@ extras = True
 [tool:pytest]
 # Options for py.test:
 # Specify command line options as you would do when invoking py.test directly.
-# e.g. --cov-report html (or xml) for html/xml output or --junitxml junit.xml
-# in order to write a coverage file that can be read by Jenkins.
+# e.g. --junitxml junit.xml in order to write a coverage file that can be read
+# by Jenkins.
 addopts =
-    --cov pyscaffold --cov-config .coveragerc --cov-report term-missing
+    --cov pyscaffold
+    --cov-config .coveragerc
+    --cov-report=
     --verbose
-#    In order to use xdist, the developer can add, for example, the following
-#    arguments:
-#    --dist=load --numprocesses=auto
+# Please call `coverage combine .coverage` explicitly after running the tests.
+# You can display/generate a report with `coverage report -m` or `coverage xml`
+#
+# In order to use xdist, please add something like
+# --dist=load --numprocesses=16
 norecursedirs =
     dist
     build
--- a/src/pyscaffold/templates/setup_cfg.template
+++ b/src/pyscaffold/templates/setup_cfg.template
@@ -57,15 +57,12 @@ extras = True
 [tool:pytest]
 # Options for py.test:
 # Specify command line options as you would do when invoking py.test directly.
-# e.g. --junitxml junit.xml in order to write a coverage file that can be read
-# by Jenkins.
+# e.g. --cov-report html (or xml) for html/xml output or --junitxml junit.xml
+# in order to write a coverage file that can be read by Jenkins.
 addopts =
     --cov ${root_pkg}
     --cov-config .coveragerc
-    --cov-report=
     --verbose
-# Please call `coverage combine .coverage` explicitly after running the tests.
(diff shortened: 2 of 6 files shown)


abravalheri:
@FlorianWilhelm if we manage to have the coverage reporting done well with tox, do you think we can avoid doing the second run before submitting to coveralls?

FlorianWilhelm:
@abravalheri, absolutely! I only introduced it because in the Travis WebGui there was no way anymore to click on the file and see which lines are covered. Thus I had to add the additional `py.test` run. I we could avoid that without too much hassle, that's cool.

setup.cfg
```diff
@@ -99,14 +99,18 @@ extras = True
...
 # Specify command line options as you would do when invoking py.test directly.
-# e.g. --cov-report html (or xml) for html/xml output or --junitxml junit.xml
-# in order to write a coverage file that can be read by Jenkins.
+# e.g. --junitxml junit.xml in order to write a coverage file that can be read
+# by Jenkins.
 addopts =
-    --cov pyscaffold --cov-config .coveragerc --cov-report term-missing
+    --cov pyscaffold
+    --cov-config .coveragerc
+    --cov-report=
```
FlorianWilhelm: Is there the `term-missing` parameter missing? Or is this preventing `pytest-cov` to do something wrongly?

src/pyscaffold/templates/setup_cfg.template
```diff
@@ -57,11 +57,15 @@ extras = True
...
-# in order to write a coverage file that can be read by Jenkins.
+# e.g. --junitxml junit.xml in order to write a coverage file that can be read
+# by Jenkins.
 addopts =
-    --cov ${root_pkg} --cov-report term-missing
+    --cov ${root_pkg}
+    --cov-config .coveragerc
+    --cov-report=
     --verbose
+# Please call `coverage combine .coverage` explicitly after running the tests.
```
FlorianWilhelm: Is this only necessary when I use `tox` to run my tests or also in the case of `py.test`?

setup.cfg
abravalheri: With `=` followed, i have tried to suppress the incorrect output (http://pytest-cov.readthedocs.io/en/latest/reporting.html)

src/pyscaffold/templates/setup_cfg.template
abravalheri: Yes, you are right :sweat_smile:

When running with `py.test` directly, the package in the source code will be used directly (in this case we need to explicitly do `python setup.py develop` or `pip install -e .[testing,all]`), so the paths will be the right ones by default (no need for `coverage combine`).

(On the other hand, when trying to run `python setup.py test` without having installed pyscaffold in development/editable mode, I get an `ImportError: module 'pyscaffold.contrib.setuptools_scm.hg' has no attribute 'FILES_COMMAND'`)

abravalheri: Since I cannot overwrite the same command line option `--cov-term` in `tox.ini` if they are specified in `setup.cfg`, my alternative proposal (a kind of trade-off that I am preparing another commit with it), would be leaving this option unspecified for `py.test`.

In this case, calling pytest directly would still print the coverage table but without the missing lines... Do you think this is a good compromise?

FlorianWilhelm: Regarding the `ImportError: module 'pyscaffold.contrib.setuptools_scm.hg' has no attribute 'FILES_COMMAND'` problem, might it be that there is still an old PyScaffold 3.0.3 version around somehow? Do you also have the problem if you start with a fresh virtualenv and clean all `.eggs` as well as `.egg-info` directories in root and `src`? This problem is caused by the fact that the `entry_points` changed in the new setuptools_scm version.

FlorianWilhelm:
@abravalheri I am not sure about this compromise. What would happen if a user adds `cov-report=term-missing` later on without knowing about this issue? Would it then suddenly stop working at all? I know you put a lot of effort into this workaround but isn't it actually a problem that one needs to fix upstream in pytest-cov? You mentioned the issue [pytest-dev/pytest-cov#130](pytest-dev/pytest-cov#130) and the PR [pytest-dev/pytest-cov#142](https://github.com/pytest-dev/pytest-cov/pull/142) seems to have solved it already and will be included in pytest-cov 2.6 as it seems. So we could keep our minimal workaround of running `py.test` again until pytest-cov 2.6 is released. What do you think?

abravalheri:
Hi @FlorianWilhelm, actually, based in the tests I have run manually, the worse that can happen is coverage.py printing the report twice in a row, the first time without merging the corresponding paths (and therefore with the wrong coverage), and the second time as one would expect (the CLI arguments are not "overridable", the first time they appear is what really counts, but no error is raised if you repeat any of them).

Basically the setup proposed in the tox file is not incompatible just very verbose, since currently pytest-cov cannot handle things internally (it would still work after the new versions, as long pytest/coverage APIs doesn't change).

But I imagine that, since it is nothing serious, we could just wait for the new release (it is just a pity that PyScaffold devs/users might see the fragmented coverage log in the terminal).

Regarding running the tests twice. That is not necessary, we could basically do:

```bash
if [[ "$COVERAGE" == "true" ]]; then (coverage combine .coverage && coveralls) || echo "failed"; fi
```

This way, despite of the report with the wrong files being printed in the terminal, coveralls would get the right one.

abravalheri:
Umm, weird. I will investigate this a little bit more. Thanks Florian

abravalheri:
Hi @FlorianWilhelm, I saw that in [.coveragerc](https://github.com/blue-yonder/pyscaffold/blob/809ac4e07a688039af8b7182383de97484900159/.coveragerc), the `[paths]` section is not configured for the tree related to this commit. I will check if with the correct setup it works.

(Maybe I was not clear in one of my previous comments, the configuration of the paths in `.coveragerc` is important for the `combine` to work properly)

abravalheri:
Here we go:

https://github.com/blue-yonder/pyscaffold/commit/8c4320b0b4bbad555b87bdd09d11bf91b59e0b56
https://travis-ci.org/blue-yonder/pyscaffold/jobs/419070168
https://coveralls.io/jobs/39531469

I think now it should be fine :)

abravalheri:
Should we close this PR?

FlorianWilhelm:
Okay, will close it.

Issue #196, opened 2018-08-13
Django extension not migrated to the `src` layout


abravalheri:
When creating a django project using PyScaffold, the created folder structure doesn't follow PyScaffold 3.* `src` layout.

### Steps to reproduce:
- With PyScaffold > 3.0, do:
```bash
$ putup myproj --django
```
- Check the generated file structure:
```bash
$ tree myproj
myproj
    ├── AUTHORS.rst
    ├── CHANGELOG.rst
    ├── docs
    │   ├── authors.rst
    │   ├── changelog.rst
    │   ├── conf.py
    │   ├── index.rst
    │   ├── license.rst
    │   ├── Makefile
    │   └── _static
    ├── LICENSE.txt
    ├── manage.py
  **├── myproj**
    │   ├── __init__.py
    │   ├── settings.py
    │   ├── urls.py
    │   └── wsgi.py
    ├── README.rst
    ├── requirements.txt
    ├── setup.cfg
    ├── setup.py
    ├── src
  **│   └── myproj**
    │       ├── __init__.py
    │       └── skeleton.py
    └── tests
        ├── conftest.py
        └── test_skeleton.py

    6 directories, 22 files
```

abravalheri:
Is this the expected behavior?
If that is the case, maybe we should add this exception to the documentation...

FlorianWilhelm:
This looks seems to be rather a bug actually. The files of `django-admin` should reside in `src/myproj`.

abravalheri:
This is also happening to cookiecutter :(

`putup myproj --cookiecutter cookiecutter-pypackage.git`

FlorianWilhelm:
I can confirm that at least for Django the problem was already in version 3.0.3. Do you know of any simple way to fix this? Maybe we need to `chdir` into `src` before we run the django command? I think for cookiecutter it's even more problematic. Here we need to change our test cookiecutter repo and also transform it into a src-layout. We will not be able to check if a cookiecutter template follows the src-layout or if it doesn't, thus we should just apply it and tell in the docs that the template should have the src-layout. For Django it's a different story, here it would be nice to somehow assert that the files are created within `src/{package}`.

FlorianWilhelm:
I forked the cookiecutter template that we use into our [PyScaffold organization](https://github.com/pyscaffold/cookiecutter-pypackage`) so that we can refer to this one in our unit tests instead of https://github.com/FlorianWilhelm/cookiecutter-pypackage.

PR #201, opened 2018-08-14
Change cookiecutter template to the src-layout.

FlorianWilhelm:
This partially solves #196 for the cookiecutter case. Now our test case:
```
putup test_project --cookiecutter https://github.com/pyscaffold/cookiecutter-pypackage.git
```
works. This was accomplished by changing the cookiecutter template to a src layout which I think is the correct way. More and more projects change to the src-layout and so will most cookiecutter templates eventually.


abravalheri:
What you think would be missing to solve #196 completely?

I completely agree with the proposal of being less magical, and requiring explicitly from the user to use a cookiecutter that adhere to the src-layout.

Issue #203, opened 2018-08-15
Bug in v3.1b2: setup_requires in setup.cfg is ignored


FlorianWilhelm:
If you run a `python setup.py develop` with a project that was setup up using PyScaffold 3.1b2 in a virtual environment where PyScaffold is not installed, the `use_pyscaffold` parameter will be ignored: `...lib/python3.6/distutils/dist.py:261: UserWarning: Unknown distribution option: 'use_pyscaffold'`

In theory, setuptools should download PyScaffold as egg into the `.eggs` directory and run it from there. Adding `setup_requires` as parameter to `setup` in `setup.py` resolves the problem. This seems to be a known [bug](https://github.com/pypa/setuptools/issues/1054).

FlorianWilhelm:
It seems having `setuptools>=38.3` installed solves the issue. So bascially we have two options:

1) Go back to have `setup_requires` as parameter in `setup.py`. This will always work but is not nice in a way.
2) Make PyScaffold itself depend on `setuptools>=38.3` and keep `setup_requires` in `setup.cfg`. This will not work if a new virtualenv with an setuptools version lesser than 38 was created and PyScaffold was not installed manually. On the other hand it's nicer and time will solve this problem since higher setuptools versions will be installed by default.

PR #205, opened 2018-08-16
Assert setuptools version

FlorianWilhelm:
This is one way to solve #203 while keeping `setup_requires` inside of `setup.cfg`.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -6,7 +6,7 @@ Requirements
 ============
 
 The installation of PyScaffold only requires a recent version of `setuptools`_,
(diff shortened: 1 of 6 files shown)


abravalheri:
Thanks Florian, seems a great solution (it never occurred to me to use `pkg_resources.working_set.add_entry` - very clever!)

FlorianWilhelm:
@abravalheri, thanks! Credits go to @ronnypfannschmidt, the maintainer of `setuptools_scm`, who gave me a little lecture on setuptools last week.

PR #209, opened 2018-08-30
Add experimental warning to pipenv

abravalheri:
I think it is good to mark this feature as experimental and state that the API is unstable (this way we have some flexibility regarding semantic versioning as well...)


FlorianWilhelm:
@abravalheri, cool, thanks!

Issue #175, opened 2018-06-13
Extract `cookiecutter` and `django` extensions into separated packages


abravalheri:
Since we don't need extensions to live inside the same code repository anymore, we could extract some of the complex ones into separated repositories.

Those two in particular already depend on extra requirements to work anyway.
So instead of doing `pip install pyscaffold[all]`, the users that want the functionalities would need to do `pip install pyscaffold pyscaffoldext-django` or `pip install pyscaffold pyscaffoldext-cookiecutter`, which is bigger, but at least we clean up the code base and speed up the tests...

Any thoughts?

FlorianWilhelm:
This seems to be great idea since it would reduce the complexity inside PyScaffold. What about potential consequences? For instance how would we do unit testing? I guess it's best to also move those to a Travis job for the plugin. What about the documentation? I think it would be great to have a small explanation of official extensions directly on our main website. 

Do you think this is something for version 3.1 or should we rather postpone it to version 3.2?

abravalheri:
Thanks for the feedback @FlorianWilhelm :)

In my view, we should move the relevant unit/system testing to the new repositories and create Travis jobs for them as well.

The main (bad) consequence that I can imagine, is having changes in PyScaffold (e.g. from PRs) that break those extensions. Right now, since the test suite is unified, any breaking change would make PyScaffold Travis job to fail, and therefore would not be merged.

If we split, we should find our way of trigger builds in the other repositories and collect the feedback before merging. Of course this is a matter of confidence in PyScaffold test suite, if the extension API is well tested, the separation would ideally not be a concern, and this downfall trigger could be neglected.

Regarding the docs, I think we should have dedicated rtds, but mention the features in summary, reference the extensions and link their docs in the main page, so it is more or less transparent for the end user.

Versions, on the other hand, are something more polemic... How PyScaffold is doing [semver](https://semver.org/) ? Are we considering versions *.0.* to be unstable in the same way semver advocates about 0.*.*? Do we consider this separation a break in the API?

(This question about semver is also related to other topics such as replacing the term "extension" by "plugin" - we can do those changes without any problem, is just that we have to plan the releases and have a deprecation period)

FlorianWilhelm:
@abravalheri I am totally in favour of semantic versioning. This is why we should wait for version 4.0 to remove those two extensions since it would break the current API. 

[...]

PR #212, opened 2018-08-31
Add deprecation notices about cookiecutter and django

abravalheri:
As discussed in #175 we should include a deprecation notice, so the users are aware in future changes in the interface.


FlorianWilhelm:
@abravalheri Thanks!

Issue #211, opened 2018-08-31
Change helpers to use pathlib.Path


abravalheri:
In the past when introducing the extensions API, one point that was controversial was the usage of either lists of file path parts or strings using the forward slash as separator for `helpers.ensure` and `helpers.reject`.

It was suggested that we could use `pathlib.Path` to simplify the code.

Now that we run on Python 3 and `pathlib` is part of the standard library we can discuss more about that.

Since it breaks the API, if we go for this approach, we would have to start deprecating it now so we can include it in the next major release.

abravalheri:
If we decide to adopt the new behaviour:

v3.2 - Deprecation Notice
v4.0 - Actual change

FlorianWilhelm:
@abravalheri Sounds like a great idea. Since it was included in Python 3.4 we should do it as you said. Thanks!

abravalheri:
Implemented in #215 with deprecation notices.
TODO note added explaining how to drop support for lists/tuples in v4.

PR #215, opened 2018-09-04
Use pathlib in helpers

abravalheri:
This PR address #211, by adding support to `os.PathLike` objects.

At this point `os.PathLike`, `str`, `list` and `tuple` are accepted (the PR just add supports, but not remove).
As a TODO for v4, we have to drop support for `list` and `tuple` .


FlorianWilhelm:
@abravalheri, nice work! Thanks a lot.

Issue #190, opened 2018-07-11
Check for clean workspace during update


FlorianWilhelm:
During an update PyScaffold could check if the current workspace really is clean. A clean workspace allows an easy rollback if an update would fail for some reason. Thus an update should only be performed when the workspace is clean. If the workspace is dirty, `putup` should emit a corresponding message and tell the user to either `commit` all current changes or `stash` the away.

PR #220, opened 2018-09-27
Check during update for clean workspace

FlorianWilhelm:
This adds a check that the Git workspace is clean during an update, so it's just a safety measure that can be overruled with `--force`. This addresses issue #190.


setup.cfg
```diff
@@ -4,7 +4,7 @@ description = Template tool for putting up the scaffold of a Python project
 author = Florian Wilhelm
 author-email = Florian.Wilhelm@gmail.com
 license = MIT
-url = https://github.com/blue-yonder/pyscaffold/
+url = https://github.com/pyscaffold/pyscaffold/
```
abravalheri: I see that the project was moved to its own organization. Neat! :+1:

src/pyscaffold/cli.py
```diff
@@ -139,10 +139,30 @@ def parse_args(args):
...
+    if opts['pretend']:
+        opts['log_level'] = logging.INFO
 
     # In case of an update read and parse setup.cfg
     if opts['update']:
-        opts = info.project(opts)
+        try:
+            opts = info.project(opts)
+        except Exception as e:
+            if opts.get('log_level', logging.ERROR) <= logging.INFO:
```
abravalheri: I think that `logger.error(..., exc_info=True)` would avoid the conditional...
But I understand that at this point we are postponing the use of `pyscaffold.log.logger` because it is just fully configured in the `pyscaffold.api.create_project` function.

Do you think it is worthy to review this design decision?

src/pyscaffold/utils.py
```diff
@@ -52,7 +52,7 @@ def chdir(path, **kwargs):
     try:
         with _chdir_logging_context(path, should_log):
             if not should_pretend:
-                os.chdir(path)
+                os.chdir(str(path))  # str to handle pathlib args
```
abravalheri: I am under the impression that in Python 3, core `os` functions where updated to automatically handle `Path`-like objects in addition to strings. I was able to run the following with no erros:

```python
import os
from pathlib import Path
os.chdir(Path('/', 'tmp'))
```

FlorianWilhelm: Interesting, for me it worked locally with Python 3.6 but it failed on Travis. Have you checked with Python 3.4? Maybe it was changed only in more recent versions.

src/pyscaffold/cli.py
FlorianWilhelm: Hmm.. good point. Have you an idea for a better resolution like moving the initialisation of the logger more upstream into cli.py? Would we lose something by doing so?

src/pyscaffold/utils.py
FlorianWilhelm: According to this [Stackoverflow question](https://stackoverflow.com/questions/41742317/how-can-i-change-directory-with-python-pathlib) it only works with Python 3.6 and above. I will add a ToDo that we drop it when we move to a required Python 3.6 version one day.

src/pyscaffold/cli.py
FlorianWilhelm: @abravalheri I moved the initialization of the logger to `cli.py` since I guess it also makes more sense to have it there. Let me know if you are fine with this and if we can merge this PR.

setup.cfg
FlorianWilhelm: Yeah, this is really cool. Blue Yonder agreed to move it over which gives us a bit more freedom in adding new contributors.

PR #224, opened 2019-01-09
first step towards "putup --single-quotes"

Laurentiu-Andronache:
I just want consistency with the double quotes, so that we might have in the future an option for:

`putup --single-quotes`

Which makes the template follow the rule:

> Use ' everywhere, except when you have a string that contains ', in which case use ".


FlorianWilhelm:
@Laurentiu-Andronache Hi, thanks for your PR. Let's see when Travis has MacOS fixed, then I'll merge. You can still push further files to this PR if you want to change the quotes also for other files.

Issue #100, opened 2017-03-24
run without installing into current environment


flip111:
The skeleton.py file does not make clear how to just run the file. Since it starts with a shebang i tried the obvious

```
chmod +x skeleton.py
./skeleton.py
```

but then i get
```
» ./skeleton.py
Traceback (most recent call last):
  File "./skeleton.py", line 24, in <module>
    from switches import __version__
ImportError: No module named switches
```
(i called my project `switches`)

this happens with python 3.5.2

I'm totally new to python, so this is really a beginner experience with this tool

FlorianWilhelm:
Hi @flip111 , 

if you check the docstring of `skeleton.py` it should read:
```
This is a skeleton file that can serve as a starting point for a Python
console script. To run this script uncomment the following line in the
entry_points section in setup.cfg:
    console_scripts =
     fibonacci = ${package}.skeleton:run
Then run `python setup.py install` which will install the command `fibonacci`
inside your current environment.
Besides console scripts, the header (i.e. until _logger...) of this file can
also be used as template for Python modules.
Note: This skeleton file can be safely removed if not needed!
```
 Do these instructions help you?

flip111:
No because

```
Then run `python setup.py install` which will install the command `fibonacci`
inside your current environment.
```

and i don't want to install commands on my system just to run this python script

FlorianWilhelm:
Are you using virtual environments `virtualenv` or conda environments btw? If you are not familiar with them you should take a look into them since they allow you to set up little isolated environments to test things like this. But take this only as a side node.

There is actually no point in running `skeleton.py` directly and you are right, I should remove the shebang :-) Will do that in the next version.

FlorianWilhelm:
Well, sorry, this was a too hasty answer. Since `skeleton.py` includes
```
if __name__ == "__main__":
    run()
``` 
this would allow people to not set a console command (by not uncomment the line in setup.cfg) but still allowing them to try this mechanism AFTER having installed it. 
Therefore what you could do, do NOT uncomment the lines 
```
    console_scripts =
     fibonacci = ${package}.skeleton:run
```
just leave them as they are und run `python setup.py install`. Then you can call the skeleton file and it will work.

flip111:
I don't get the point of this install. Why do i need the extra step? What is being installed? What does it give me? Where are the files installed? Can i also uninstall?

[...]

PR #225, opened 2019-02-03
Remove unnecessary shebangs.

QuLogic:
* `conftest.py` is auto-loaded by pytest; it was never intended to be run.
* `setup.py` should never be run directly. The recommended command is
  `pip install .`.
* `skeleton.py` is meant to be run via setuptools' `console_scripts`
  option (see #100, for example)
* `test_skeleton.py` contains no main code and is meant to be run via
  pytest, not the interpreter directly.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -57,7 +57,7 @@ md =
 testing =
     sphinx  # required for system tests
     flake8  # required for system tests
-    pytest
+    pytest<4.0.0
     pytest-cov
     pytest-virtualenv
     pytest-xdist
--- a/setup.cfg
+++ b/setup.cfg
@@ -61,6 +61,7 @@ testing =
     pytest-cov
     pytest-virtualenv
     pytest-xdist
+    virtualenv
     # We keep pytest-xdist in the test dependencies, so the developer can
     # easily opt-in for distributed tests by adding, for example, the `-n 15`
     # arguments in the command-line.
--- a/src/pyscaffold/templates/conftest_py.template
+++ b/src/pyscaffold/templates/conftest_py.template
@@ -1,4 +1,3 @@
-#!/usr/bin/env python
 # -*- coding: utf-8 -*-
 """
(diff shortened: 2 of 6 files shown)


FlorianWilhelm:
Hi, thanks. You are right about most points except of `setup.py` which can be run directly but most people use `python setup.py` anyway, so it's okay to remove the shebangs.

Seems like the unit tests are broken due to an py.test upgrade while the virtualenv plugin still needs pytest<4.0.0. Can you fix that temporarily? Then I will merge your PR. Thanks

FlorianWilhelm:
Sorry, my bad, looking back in the other build logs, I think it needs to be pinned to `pytest<4.2`. Thanks!

QuLogic:
No, it's definitely `<4.0.0` from the logs. I've added `virtualenv` to the requirements too as it seems `pytest-virtualenv` doesn't depend on it itself.

FlorianWilhelm:
Strange that only Python 3.4 in conda fails now. I am on vacation right now without good internet access. Have to look into it in a week or so.

QuLogic:
Is there any reason to continue supporting 3.4? It's in "security fixes only" mode, and not even releasing binary installers.

FlorianWilhelm:
@QuLogic Good point, if you want to you can remove the test and increase the corresponding minimum Python version to 3.5 in `setup.cfg`. I think some changes also need to be done to the documentation which mentions 3.4 sometimes. 

Right now I have not much spare time time to work on PyScaffold. Thanks for your support, I really appreciate it.

FlorianWilhelm:
One could remove the unnecessary shebangs also not only in the templates but also in the source code of PyScaffold itself wherever it makes sense.

FlorianWilhelm:
@QuLogic Thanks for your PR. I merged and will try to fix the problem with Python 3.4. Looking forward to more PRs :-)

PR #234, opened 2019-06-14
Check for environment vars for git config

noahp:
Before checking for global git-config `author` and `name` entries, check
for the git environment variables.

Git will use these environment variables when creating commits, falling
back on git-config values if they aren't set:
https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables#_committing

Using the environment variables instead of a global git-config setting
is useful when dealing with repositories with different author settings
(eg https://knpw.rs/blog/multiple-git-users ).

### DIFF
--- a/src/pyscaffold/info.py
+++ b/src/pyscaffold/info.py
@@ -7,6 +7,7 @@ import copy
 import getpass
 import socket
 from operator import itemgetter
+import os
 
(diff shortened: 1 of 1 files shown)


FlorianWilhelm:
@noahp Thanks a lot for this PR. Really useful feature!

Issue #238, opened 2019-07-15
test_install.py failing and virtualenv issues


abravalheri:
(Related to  #119)

The following command is currently failing:

```bash
$ tox -e default -- tests/test_install.py

# ...

tests/test_install.py::test_sdist_install FAILED                         [  6%]
tests/test_install.py::test_sdist_install_dirty FAILED                   [ 12%]
tests/test_install.py::test_sdist_install_with_1_0_tag FAILED            [ 18%]
tests/test_install.py::test_sdist_install_with_1_0_tag_dirty FAILED      [ 25%]
tests/test_install.py::test_bdist_install FAILED                         [ 31%]
tests/test_install.py::test_bdist_wheel_install FAILED                   [ 37%]
tests/test_install.py::test_git_repo FAILED                              [ 43%]
tests/test_install.py::test_git_repo_dirty FAILED                        [ 50%]
tests/test_install.py::test_git_repo_with_1_0_tag FAILED                 [ 56%]
tests/test_install.py::test_git_repo_with_1_0_tag_dirty FAILED           [ 62%]
tests/test_install.py::test_sdist_install_with_data FAILED               [ 68%]
tests/test_install.py::test_bdist_install_with_data FAILED               [ 75%]
tests/test_install.py::test_bdist_wheel_install_with_data FAILED         [ 81%]
tests/test_install.py::test_edit_install_with_data FAILED                [ 87%]
tests/test_install.py::test_setup_py_install_with_data FAILED            [ 93%]
tests/test_install.py::test_setup_py_develop_with_data FAILED            [100%]
```

### Steps to reproduce:
1. Clone PyScaffold v4.0.x branch
2. Comment the extra dependencies for `all` in `setup.cfg` (required since extensions are not updated yet.
3. Run tox

### Versions and main components
* PyScaffold Version: v4.0.x head of the branch
* Python Version: 3.6.8
* Operating system: Ubuntu
* How did you install PyScaffold: pipx

### Meaningful pieces of trace:
```
_______________________ test_sdist_install_with_1_0_tag ________________________

demoapp = <tests.test_install.DemoApp object at 0x7f26e0eaa550>

    def test_sdist_install_with_1_0_tag(demoapp):
        (demoapp
            .make_dirty_tree()
            .make_commit()
            .tag('v1.0', 'final release')
>           .build('sdist')
            .install())

$HOME/pyscaffold/tests/test_install.py:234: 
[...]

PR #239, opened 2019-07-15
Fix tests that require isolation via virtualenv

abravalheri:
This PR attempts to solve #238.
In order to do so, it replaces pytest-virtualenv dependency by custom code based on stdlib's venv.
To workaround the apparent inability of venv to install pip when running from inside another venv, we use the global python3 available.


.travis.yml
```diff
@@ -7,13 +7,17 @@ matrix:
   fast_finish: true
   include:
     - os: linux
+      dist: xenial
```
abravalheri: For some reason, travis insists in using an old version of ubuntu, and this version have problems using apt (gpg keys issue)

abravalheri:
I think I am happy with the code now. Sorry if I took me a while, every time we have a Travis issue, the process in painfully iterative.

To be fair, there are a few more things that this PR do:
- Switch to Travis built-in way of installing packages instead of manually using `apt` or `brew`
- Remove duplication for the system tests (e.g., before: a lot of `run` functions around, now: everything goes to the same `run`) and centralize helper functions/fixtures in a single place.
- Remove test dependencies on `pytest-shutils` and `pytest-virtualenv`
- Use stdlib's `venv` module to isolate tests
- Better `run` helper (with better support to `env` and `verbose`
- Remove workaround needed to make `Pipenv` (and other dotenv programs) work with `pytest-virtualenv`
- Remove tests to ensure `requirements.txt` is installed.
- Remove some unnecessary things since now we are using Python > 3.4

Unfortunately this PR also removes the ability of using coverage inside the venvs used during tests.

(P.S.: The falling test in Travis happens because we are trying to use Python 3.5 but we require 3.6 minimum)

FlorianWilhelm:
@abravalheri, thanks for this PR. I ported most of it to version 3.2.3 and the current v4.0.x branch. I stayed with pytest-virtualenv and it seems to works. Let me know if you want to have something else merged. 

Going to Cirrus was a great idea btw. My suggestion would be to have a version 3.3 first with `--cirrus` and also update `pyscaffoldext-custom-extension` to use that instead of `--travis`. Also, all current extension should be updated first. After that, we could go on working on 4.0. 

You can just close this PR if you think everything is ported.

Issue #240, opened 2019-07-17
Review PyScaffold's own setup.cfg


abravalheri:
For example, there are classifiers for old versions of Python there.

There is also a classifier for Windows, although our CI doesn't run any Windows check.

PR #242, opened 2019-07-18
Address issue #240 by removing some classifiers

abravalheri:
… in `setup.cfg`, related to old versions of Python.

Moreover, it removes explicity Windows support, since we don't have any
CI in place targetting Windows.

(#240)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -18,8 +18,6 @@ classifiers =
     Programming Language :: Python
     Programming Language :: Python :: 3
     Programming Language :: Python :: 3 :: Only
-    Programming Language :: Python :: 3.4
-    Programming Language :: Python :: 3.5
     Programming Language :: Python :: 3.6
     Programming Language :: Python :: 3.7
     Environment :: Console


FlorianWilhelm:
I hope to have some time to fix the versions of our extensions on Saturday to make our unit tests run again. Regarding the Windows support, I fear that we might lose some of our users with the new release then. We have special codepaths for windows and I test it sometimes manually on a Windows machine. We should consider though using an additional service like Travis or maybe even replace it with something that is also for free and supports Linux, Windows and Mac. Do you have an idea?

FlorianWilhelm:
Should we go then for cirrus-ci if it supports everything and drop Travis? Do you think it's much effort. Last thing I heard is that Travis also fired a lot of their employees.

abravalheri:
Let's try it!

FlorianWilhelm:
@abravalheri, I tried bumping the version of all extensions to some alpha versions with the corresponding `pyscaffold>=4.0a0,<4.1a0` settings but the PyScaffold unit tests still don't work since pip doesn't seem to install development versions when the `[all]` qualifier is used although PyScaffold 4.0 is also a development version... welcome to dependency hell. Having non-dev extension versions for PyScaffold 4.0 would be dangerous since users might get confused. I think the most practical thing to do is just to remove the `[all]` for the time being on the `v4.0.x` branch.

FlorianWilhelm:
@abravalheri, regarding Cirrus CI: I set our Github repo up to use it and now one would need to adjust `.cirrus.yml` to reflect the same logic as in Travis I guess and see how Linux, Mac and Windows can be integrated into one matrix. Sorry that I can be of not that much help right now, but I got a lot on my plate at work right now and can only contribute so much to PyScaffold right now.

abravalheri:
Hi @FlorianWilhelm, just a side note here:

I finally managed to put some half-decent configuration for Cirrus-CI in the `add-cirrus-ci` branch.
Of course, some things are missing (like: caching, `tests/system/test_dependency_managers.py::test_pipenv_works_with_pyscaffold` is failing, windows is failing), but I am impressed that it runs much faster. I think now we are taking full advantage of `pytest-xdist` (and the fact that the containers have the right version of python already pre-installed help a lot).

Regarding the extensions. I think the best for now is focus in the main repo, when we are good to go, we release an alpha version of PyScaffold, and only then start working in a v4.0.x branch for the extensions. It is not worthy to be super concerned with them right now...

abravalheri:
Since #243 introduces proper automated testing on Windows, I included the Windows classifiers back.
Although we still have some errors happening, I think it is better to discuss those in a different issue.

I will merge this PR for now, since all the polemic bits are gone :)

PR #243, opened 2019-07-25
Add CI tasks for Windows environments through Cirrus CI

abravalheri:
This PR adds Windows testing though Cirrus CI.

Additionally, it also fixes some of the code in order to make it work better on Windows / other versions of Python / other environments.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,25 +1,35 @@
+---
+# Default values to be merged into tasks:
 auto_cancellation: false
-environment:
+env:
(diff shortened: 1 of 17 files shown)


abravalheri:
@FlorianWilhelm, now that Cirrus CI is in place covering Linux, OSX and Windows, should we remove Travis or keep both CIs?

FlorianWilhelm:
@abravalheri thanks a lot for setting this up. I think keeping only one CI makes total sense, thus we can drop Travis. Do you know why the Cirrus badge keeps saying "Build unknown"?

PR #246, opened 2019-07-28
Potential fix for #238

FlorianWilhelm:
Hi @abravalheri, would this solve the issue with pytest-virtualenv? According to the output when running `tox -- -k test_sdist_install_with_1_0_tag`:
```
Already using interpreter /Users/fwilhelm/Sources/pyscaffold/.tox/default/bin/python
Using real prefix '/Users/fwilhelm/anaconda/envs/pyscaffold'
New python executable in /var/folders/s7/6tn210s12s524j8w7vftx5mw39526l/T/tmp8d2d9kfm/.env/bin/python
Installing setuptools, pip, wheel...
```
it seems like the correct Python interpreter from Tox is used. Would this solve #238?

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -57,6 +57,7 @@ ds =
     pyscaffoldext-dsproject
 # Add here test dependencies (used by tox and pytest-runner)
 testing =
+    ipython
     sphinx  # required for system tests
     flake8  # required for system tests
     pytest
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,6 +4,7 @@ import logging
 import os
 import shlex
 import stat
+import sys
 from collections import namedtuple
 from contextlib import contextmanager
 from importlib import reload
@@ -13,6 +14,8 @@ from shutil import rmtree
 from pkg_resources import DistributionNotFound
 
 import pytest
+
+from pytest_virtualenv import VirtualEnv
 
 from .helpers import uniqstr
 
@@ -44,8 +47,9 @@ def command_exception(content):
 
 
 @pytest.fixture
-def venv(virtualenv):
+def venv():
     """Create a virtualenv for each test"""
+    virtualenv = VirtualEnv(python=sys.executable)
     return virtualenv
 
 


abravalheri:
Hi @FlorianWilhelm, this is exactly the other approach I had in mind.

I should be OK, but I don't know. With `venv` I had problems even when using `sys.executable`, because it was pointing to a symlink inside an existing venv... Every time venv was not created with a root installation of Python, I had problems (it was re-using packages from the "parent" venv)... But maybe virtualenv is smarter then venv...

Additionally the Cirrus CI logs say:
```
pytest_shutil.workspace:workspace.py:112 run: ['c:\\users\\containeradministrator\\appdata\\local\\temp\\cirrus-ci-build\\.tox\\all\\scripts\\python.exe', '-m', 'virtualenv', '-p', 'c:\\users\\containeradministrator\\appdata\\local\\temp\\cirrus-ci-build\\.tox\\all\\scripts\\python.exe', 'C:\\Users\\ContainerAdministrator\\AppData\\Local\\Temp\\tmp8xudg1ws\\.env']
```
which is a bit odd...

I think the best approach to test if this fix is valid is to port back the changes from #239 and see if the updated test suite passes... Sorry if I put lot's of code there, I was working to make venv work ad at the same time fixing windows bugs and at the same time finding problems of isolation with the venv and bdist ... a mess

abravalheri:
Hi @FlorianWilhelm, I basically took the updated test suite from #239 (that, right now is basically "green": https://cirrus-ci.com/build/5748010640736256 - just the coverage is not being uploaded - I will correct the `.cirrus.yml` file when we finish fixing this issue) + the workarounds for windows and merged with the concepts you proposed in this PR (using `pytest_virtualenv`).

Hopefully the output of the Cirrus CI will not be cryptic and will tell if this approach works.

FlorianWilhelm:
@abravalheri, thanks a lot! I just reviewed your PR #247.

abravalheri:
More or less. I did it in a weird way to "preserve the semantics" of the PRs :P.

Basically we have #239 | (#247 & #246).

#247 augments #246. So i should be merged first. Then #246 can be merged.

Or I edit #247 and simplify stuff.

However we still have weird/inexplicable windows errors on #247 that don't occur in #239.

FlorianWilhelm:
Hi @abravalheri, I know that you already put much effort into those three PRs, but do you think it might help us to split them up some more? For instance, the necessary changes to just move from Travis to CirrusCI could be done in a PR. Without any Windows support, just what we have but on CirrusCI. In general, I feel like smaller PRs which focus on one little step would help us to keep going forward and not stuck after a while. Especially for me, it's easier to review such a small PR.  What do you think? We could merge faster and speed things up.

FlorianWilhelm:
This PR is no longer necessary and the actual changes are already in master.

Issue #244, opened 2019-07-25
Some tests failing on Windows


abravalheri:
After the discussion in #242, it is evident that Windows support is important for PyScaffold.
#243 address this topic by adding automated tests on Windows via Cirrus CI. However, as indicated in one of the [Cirrus CI builds](https://cirrus-ci.com/task/5836154039959552), some tests are still failing on Windows (despite of a huge effort I put into debugging it 😢)

- [test_info.py:19](https://cirrus-ci.com/task/5836154039959552?command=test#L632) error seems to be just silly, and I don't understand why it is happening.
  According to [Python docs](https://docs.python.org/3.5/library/getpass.html), if `USERNAME` is set, we shouldn't be seeing this error... But we are.
  This is not super important since the user is supposed to configure git anyway...
  So I am wondering, should we skip this test on Windows using PyTest marks?

- [test_repo.py:74](https://cirrus-ci.com/task/5836154039959552?command=test#L1299) error is enigmatic...
  I have 0 clues why we don't have access rights to the git repository.
  If we can create the folder, we should be able to remove it...
  But again, it might not be super important... Looks like it is a super specific sequence of actions in the CI env that is causing that...

- [test_install.py:123](https://cirrus-ci.com/task/5836154039959552?command=test#L3980) error is laughable...
  `The filename or extension is too long` is a very poor error message for the fact that Microsoft hardcoded a 260 char limit for file names and never bothered correcting it...
  [It doesn't seem to be easy to solve](https://www.howtogeek.com/266621/how-to-make-windows-10-accept-file-paths-over-260-characters/), but yet again not super important.
  It is an issue with the test suite, not necessarily PyScaffold itself.

The remaining error messages seem to be related, at first glance, to #238 (problems with virtualenvs - particularly testing `bdist`s seems to be a recurring pain... to the first time we have problems with that test case).

@FlorianWilhelm do you have any thoughts on this matter, or any approach for dealing with the errors on Windows?

FlorianWilhelm:
Hi @abravalheri, thanks a lot for your debugging efforts. Regarding your points:

> test_info.py:19 error seems to be just silly, and I don't understand why it is happening. According to Python docs, if USERNAME is set, we shouldn't be seeing this error... But we are. This is not super important since the user is supposed to configure git anyway... So I am wondering, should we skip this test on Windows using PyTest marks?

[...]

PR #247, opened 2019-07-28
Update #246 with changes from #239

abravalheri:
This PR aims to bring together the proposals in #246 and #239.

As mentioned in #246, one way of checking that the approach works is by running the "updated" test suite from #239. This PR aims to make that possible (by merging the other 2)

Changes include:
-  Moving helpers for system tests (including virtualenv logic) previous scattered into `tests/systems/__init__.py` (e.g. the equivalent of the previous functions `run`, `venv_run`, etc...)
- Subclassing `VirtualEnv` (so we can trigger `pip`, `setuptools` and `certifi` installations if necessary after the instantiation, and therefore avoid major problems when testing in 3.5 and Windows). Also allows for the logic/workarounds previous in `venv_run` to be embedded in the venv object itself.
- Loads of changes to make test "pass" (or at least xfail) on Windows, as described in #244 .

Hopefully the CI tests will pass and we will see that the approach in #246 is valid.


tests/system/__init__.py
```diff
@@ -0,0 +1,176 @@
...
+class Venv(VirtualEnv):
+    """High-level interface for interacting with :mod:`venv` module.
+
+    Arguments
+        path(os.PathLike): path where the venv should be created
+    """
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        for k, v in environ.items():
+            if 'PIP_' in k:
```
abravalheri: Hopefully this will allow us to use PIP_TRUSTED_HOST at least once, when updating `pip/setuptools/certifi`.
From there on, `setuptools` should be able to use `certifi` up-to-date certificates when installing build dependencies listed in `setup_requires` (Related to SSL errors on Windows, described in #244)

tests/system/__init__.py
```diff
@@ -0,0 +1,176 @@
...
+                self.env[k] = v
+        self.path = self.virtualenv
+        self.bin_path = self.path / BIN
+        self.pip_exe = which('pip', path=str(self.bin_path))
+        self.python_exe = which('python', path=str(self.bin_path))
+        assert self.python_exe and self.pip_exe
+
+        if self.pip_version() < MIN_PIP_VERSION or IS_WIN:
+            self.run(self.python_exe, '-m', 'pip', 'install', '--upgrade',
+                     'pip', 'setuptools', 'certifi')
```
abravalheri: This is a workaround for a few problems:
- old installations of `setuptools` will not be able to recognize `use_pyscaffold` keyword in `setup.py`
- SSL errors in windows (see #244)

tests/system/__init__.py
```diff
@@ -0,0 +1,176 @@
...
+            # ^  this makes tests slower, so try to avoid it
+            #    certifi: attempt to solve SSL errors on Windows
+
+    def run(self, *args, **kwargs):
+        kwargs.setdefault('cd', getcwd())
```
abravalheri: `pytest_virtualenv` (that uses `pytest_shutil`), is very weird when running commands...
The `cwd` might be different from  what is expected, so we ensure expectations are aligned.

tests/test_update.py
```diff
@@ -134,7 +131,8 @@ def putup(self, *args, with_coverage=False, **kwargs):
         if with_coverage:
             # need to pass here as list since its args to coverage.py
             args = [subarg for arg in args for subarg in arg.split()]
-            putup_path = path_join(self.venv_path, 'bin', 'putup')
```
abravalheri: This won't work on windows... So I replaced it with `which`

FlorianWilhelm: Nice, didn't know about `which` being available in `shutil`.

tests/system/__init__.py
```diff
@@ -0,0 +1,176 @@
...
+    To workaround this we can always prepend the command with `env`, this
+    guarantees environment variables have up-to-date values.
+    """
+    if len(args) > 0 and args[0] not in ('env', ENV):
+        args.insert(0, ENV)
+
+    return args
+
+
+def run(*args, **kwargs):
```
FlorianWilhelm: I wonder if we have this kind of functionality already in `pyscaffold.shell.ShellCommand` or if we could at least make use of it to avoid code duplication here.

tests/system/__init__.py
FlorianWilhelm: Isn't the installation of `pip` and `setuptools` in recent versions not something we should do within `tox.ini` or `.cirrus.yml`? Ah, wait, we are within a new virtualenv and for that reason we have to do this, right? Okay, fine then :-)

tests/system/test_install.py
```diff
@@ -7,18 +7,18 @@
  C: a commit after a tag, clean tree
  D: a commit after a tag, dirty tree
 
-The tests written in this file use pytest-virtualenv to achieve isolation.
+The tests written in this file use venv to achieve isolation.
```
FlorianWilhelm: Is this right? I thought we are now still using pytest-virtualenv but wrapping it with some helpers in `Venv`?

tests/test_update.py
```diff
@@ -164,27 +162,25 @@ def venv_mgr(tmpdir, venv, pytestconfig):
 
 @pytest.mark.slow
 def test_update_version_3_0_to_3_1(venv_mgr):
-    project = path_join(venv_mgr.venv_path, 'my_old_project')
+    project = venv_mgr.venv_path / 'my_old_project'
```
FlorianWilhelm: Nice! We should make match more use of pathlib now that we use Python >=3.5/6.

tests/system/__init__.py
abravalheri: Exactly, it a bit painful and repetitive... Unfortunately, even if we update `setuptools` and `pip` in the "parent" python installation, these changes will not propagate to the the "derivative" virtualenv. The derivative virtualenv will contain the initial versions that were installed in the machine, and therefore we will have to install the new ones manually...

This is the price to pay for attempting to write system-level tests in python 😢 . Loads of meta-thinking. We are using one "installation" of python to drive another "installation".

I was wondering if we can use something like [nox](https://nox.thea.codes/en/stable/) to try to eliminate some layers in this process (right now we have `tox > pytest > subprocess [> pytest]` -- `>` meaning "driving" and `[]` meaning "in some cases" -- maybe with `nox` we could have `nox > subprocess [> pytest]`, since `nox` itself is scriptable and have an API similar to `pytest`), but I guess that should be experimented in an extension first (when I have the time), to see if it makes sense...

tests/test_update.py
abravalheri: Yes, I discovered it last week when messing around with this issue. I guess it is a wonderful Python 3 addition.

Python's stdlib always surprises me.

tests/system/test_install.py
abravalheri: Sorry, I did a quick merge between the branches and corrected the errors, but I forgot to come back correcting the documentation.

You are completely right. I have to change this comment.

abravalheri:
Hi @FlorianWilhelm, thank you for the review.

Unfortunately we have *yet another **new** cryptic error for Windows*™️ after the merge of the PRs that is not present in #239, and I don't know how to explain. So I am not comfortable with merging yet.

```
FileNotFoundError: [Errno 2] No such file or directory: 'c:\\users\\containeradministrator\\appdata\\local\\temp\\cirrus-ci-build\\.tox\\all\\lib\\site-packages\\pyscaffold\\templates\\gitignore.template'
```

It is like `putup` was not doing its job...

For fixing that I have to include more debugging messages, and this might take some time (working solely with the CI output is very painful because it is so slow).

On top of that, I am not exactly sure about coverage though... If we read `pytest_shutil`'s docstring for `run_with_coverage`, we can see the following:
```
    cd: `str`
        If not None, will change to this directory before running the cmd.
        This is the directory that the coverage files will be created in.
```
And we have being using `cd` (since before this PR), to guarantee the tests behave in an intuitive way. We `cd` into `cwd` before running the commands, so for example `putup` works as expected (it has a path dependent argument)... So I am wondering now if we don't leave a bunch of coverage files left behind inside a `tmpdir`, that are not merged when we run `coverage` at the end of the tests...

In #239's [coverage-run](https://github.com/pyscaffold/pyscaffold/pull/239/files#diff-f807ddb74cf249061e9dbca36c0c5b94R213) function,  I tried to workaround that subject by copying the files back to the main project folder (no idea if it works or not, though, since the Cirrus CI task for publishing the coverage is broken 😝)

FlorianWilhelm:
Hi @abravalheri, hm... that's odd. It looks like the installation of PyScaffold went wrong when `site-packages\\pyscaffold\\templates\\gitignore.template` is missing. This didn't happen before, right? So strange, why does all of that happen only for Windows?

Regarding the coverage with pytest-virtualenv, I am pretty sure that this works on the master branch for PyScaffold 3.2. I remember that it took me quite a while to get all settings right so that these tests are also part of the coverage report.

FlorianWilhelm:
@abravalheri, thanks for this PR. I ported most of it to version 3.2.3 and the current v4.0.x branch. I stayed with pytest-virtualenv and it seems to works. Let me know if you want to have something else merged. 

Going to Cirrus was a great idea btw. My suggestion would be to have a version 3.3 first with `--cirrus` and also update `pyscaffoldext-custom-extension` to use that instead of `--travis`. Also, all current extension should be updated first. After that, we could go on working on 4.0. 

You can just close this PR if you think everything is ported.

PR #253, opened 2019-10-23
Python 3.8 on Linux and Windows

jayvdb:
The Windows choco package has been updated to use Python 3.8.
Lots of users will also been switching to Python 3.8 as a result, so adjust CI to continue using the default choco package.


FlorianWilhelm:
@jayvdb Thanks a lot for your PR!

PR #254, opened 2019-10-23
Use sys.executable in tests

jayvdb:
This is incomplete.

The motivation is packaging of PyScaffold for linux distros, and running the test suite outside of a venv where `python` may not exist at all, or points to `python2` instead of `python3` which causes problems when running the test suite on python3.

for reference, my project to package PyScaffold is at https://build.opensuse.org/package/show/home:jayvdb:PyScaffold/python-PyScaffold -- the .spec has some hacks in it which this patch is trying to tidy up.

### DIFF
--- a/tests/system/helpers.py
+++ b/tests/system/helpers.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import shlex
+import sys
 from os import environ
 from subprocess import STDOUT, check_output
 
@@ -11,6 +12,9 @@ def run(*args, **kwargs):
             args = shlex.split(args[0])
         else:
             args = args[0]
+
+    if args[0] == 'python' and sys.platform != 'win32':
+        args[0] += str(sys.version_info[0])
 
     opts = dict(stderr=STDOUT, universal_newlines=True)
     opts.update(kwargs)
--- a/tests/test_repo.py
+++ b/tests/test_repo.py
@@ -2,6 +2,7 @@
 import os.path
 import shutil
 import subprocess
+import sys
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


abravalheri:
Hi @jayvdb thanks very much for the PR :)

I have a question regarding using `sys.executable`:

If I remember well from our previous experiments (for more information please have a look on #238 #239 #246 #247), it is tricky to use `sys.executable` combined with virtualenv/venv [[1]](https://virtualenv.pypa.io/en/stable/reference/#compatibility-with-the-stdlib-venv-module). Will this change affect the system tests/break their isolation (since they are based in "nested" venvs)?

jayvdb:
The PR title "use `sys.executable`" is a bit misleading.  The problem this PR is fixing is use of `'python'` in the *top-level* tests suite, which should be able to run outside of tox, and isnt happening in any nested VM.

For one commit, I am replacing string `'python'` with string `'python2'` or `'python3'`.

The other commit uses `sys.executable`, and it seems quite safe to do so as it is only prep of the demo app.  But if you prefer, I can change that commit to use `'python2'`/`'python3'` as that also works for me.

I also have some changes needed for invocation within the venvs, but they are trickier as you know, so this PR is the easier ones.  Currently I only have hacks for this, not proper patches.  I'll review your existing PRs before going deeper, as you may already have solved that set of problems.

FlorianWilhelm:
@jayvdb lgtm, do you want to add more since it has still WIP in the title or should I merge?

jayvdb:
Rebased and removed WIP: prefix.  Thanks for checking.

FlorianWilhelm:
@jayvdb Thank you!

Issue #138, opened 2017-12-28
Allow inplace updates


FlorianWilhelm:
In order to update a PyScaffold project it is necessary to `cd` into its parent directory and issue a `putup --update my_project` command. It would be more convenient to allow `putup --update .` from within the project root directory.

FlorianWilhelm:
Let's postpone this to version 3.2

FlorianWilhelm:
Postponed to version 4.0 due to necessary API changes for this feature to be implemented cleanly. Check out #219 for more details on this.

abravalheri:
Just a brainstorm for the `--update` flag.

For the far future, it should be a built-in extension :P to allow something like:
```bash
putup . --update strategy:git-merge
putup . --update strategy:keep
putup . --update strategy:ask  # To be implemented by an "Interactive" extension
putup . --update   # By default uses "strategy:builtin"
```
- `strategy:git-merge`: PyScaffold could create a branch when updating, replace all the files and then attempt a merge. If the merge fails the user have to solve a merge conflict. Very straight forward
- `strategy:keep`: PyScaffold can create a `<filename>.alt~` file in the case of conflict, so the user can merge the files manually
- `strategy:builtin`: PyScaffold uses the `FileOp` metadata and some heuristics like it already does nowadays.

Just nice to have in mind.

FlorianWilhelm:
It's a cool idea that would give us much more flexibility but also complicate things a lot. For me the following workflow always worked.
1. Use `putup --force ...`
2. Run `git difftool` with `meld`

I think much of the functionality is already covered by git itself. Wouldn't you agree?

abravalheri:
Hi @FlorianWilhelm, you are right. I haven't thought about this workflow.
Maybe we should create a section in the docs for that.

By the way, I think this issue is mostly closed. However, it is still necessary to create a test case for it.
For #241, I managed to to test in my computer by locally merging #239, but I prefer waiting for the tests to be fixed before touching `test_update`, since it is heavily changed by #239 (in the case the PR is abandoned).

abravalheri:
@FlorianWilhelm is there any other test you would like to do before closing this topic?

FlorianWilhelm:
Looks good to me! Thanks, @abravalheri!

PR #262, opened 2020-01-13
Add tests for inplace update (#138)

abravalheri:
This PR adds a test as described in #138.

@FlorianWilhelm, the test is very simple. Please let me know if you have any other ideas on what should be tested.

I also have a doubt:
- Right now if I first start the project with no description and I try to update the project later using the `-d/--description` flag, the given text is not update in `setup.cfg`.
Is this the expected behaviour?


FlorianWilhelm:
Thanks a lot @abravalheri.

I think it is updated if you use `--force`. The idea was that you might have own changes in `setup.cfg` already. I am also not sure if many people would use another `putup` call to make a change like that.

PR #263, opened 2020-01-14
Rework get_template to be used in extensions

### LINKED ISSUE #252: Rework get_template to be used in extensions
Provide a version of `get_template` that can be used in extensions to avoid code duplication.

abravalheri:
This commit closes #252

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -39,7 +39,7 @@ regular_task_template: &REGULAR_TASK_TEMPLATE
   prepare_script: &prepare_script
     - git config --global user.email "you@example.com"
     - git config --global user.name "Your Name"
-    - rm -rf .coverage junit-*.xm5
+    - rm -rf .coverage junit-*.xml
     # ^  avoid information carried from one run to the other
   test_script:
     - python setup.py egg_info --egg-base .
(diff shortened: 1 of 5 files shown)


FlorianWilhelm:
This looks great, thanks @abravalheri! Regarding the failed PY3.5 check. Should we pin `pipx` to a version that runs with PY3.5? At least for a PyScaffold version 3.3. With version 4.0 we will drop the PY3.5 support.

abravalheri:
Hi @FlorianWilhelm, I think that is sensible, otherwise it will break all the builds in the CI.
It is a pity `pipx` moved away from Python 3.5 before 2020-09-13. There are a few important distributions that I guess will rely on 3.5 up to that date (some Ubuntu LTS for example).

abravalheri:
And we see ourselves again in the middle of a Dependency Hell 😝 

Apparently `pipx==01.0` solves 3.5 but breaks other tasks.

FlorianWilhelm:
Hi, is there anyone still working on this?

abravalheri:
Hi @FlorianWilhelm regarding the `get_template` itself, I think this PR solves the issue.
However the CI is complicated... and I don't have many ideas on how to start solving it.

abravalheri:
Hi @FlorianWilhelm, I managed to make the tests start running again on windows and osx.
I believe the best would be merge this change (since the tests show the changed function works even for windows).
If there is another problem with the CI, the best would be try to address them in a separated issue.

FlorianWilhelm:
Thanks a lot @abravalheri!

PR #273, opened 2020-03-29
Debianize

Elbandi:
Create debian package


FlorianWilhelm:
@Elbandi Thanks a lot!

PR #274, opened 2020-04-14
Double quote $HOME in  travis_install.template

johanvdw:
Double quote to prevent globbing and word splitting
https://github.com/koalaman/shellcheck/wiki/SC2086


FlorianWilhelm:
Thank you!

PR #283, opened 2020-07-06
Deprecate pytest-runner and sphinx integration to setup.py

abravalheri:
As much as could find, I changed the documentation details.
Additionally, the tox template is taking the responsibility for the old commands.


FlorianWilhelm:
Thanks

PR #284, opened 2020-07-07
Add Cirrus CI built-in extension

### LINKED ISSUE #251: Add Cirrus CI - Flag for core Pyscaffold
Travis does not support Windows builds but Cirrus does. PyScaffold is migrated to using Cirrus CI now instead of Travis and thus we should also have a `--cirrus` flag to make it easy for users to set up their projects with Cirrus CI.

A nice `.cirrus.yml` template can be derived from [configupdater](https://github.com/pyscaffold/configupdater/blob/master/.cirrus.yml).

abravalheri:
The original template was taken from configupdater.

The extension automatically activates pre-commit and tox (even if
developer does not use them locally, they still are a good addition to
any CI environment).

This commit closes #251


FlorianWilhelm:
Thanks!

PR #289, opened 2020-07-09
Introduce global config

### LINKED ISSUE #236: Introduce global config
Since PyScaffold users probably use `putup` to create new packages in a recurring basis, it would be nice to allow them to "save" their favourite options somewhere.

### Proposal
The `create_project` function could deep merge the arguments coming from the current sources:
1. Global configuration file (or file specified by an `--config PATH` flag)
2. Local `setup.cfg` (or maybe `pyproject.toml` like everyone seems to be doing nowadays?)
3. Given arguments (via cli or direct api call)
(in order)

Additionally a flag `--save-config` could be used in any `putup` invocation to save the given parameters to the global config file (to *bootstrap* the config file creation).

### Global Config File:
Something following what PyScaffold already does with per-project `setup.cfg`, but in a global location:
- *OSX*: `~/Library/Preferences/pyscaffold/default.cfg` (or something like that)
- *Windows*: `C:\Users\<username>\AppData\Local\pyscaffold\pyscaffold\default.cfg`
- *Linux/Others*: `$XDG_CONFIG_HOME/pyscaffold/default.cfg` with fallback to `~/.config/pyscaffold/default.cfg`
(Please check [`appdirs`](https://pypi.org/project/appdirs/) for more info).

abravalheri:
As discussed in #236, this commit introduces the ability of storing your
"favourite" configuration for PyScaffold in a config file in the user's
home directory:

- OSX: ~/Library/Preferences/pyscaffold/default.cfg (or something like that)
- Windows: C:\Users\<username>\AppData\Local\pyscaffold\pyscaffold\default.cfg
- Linux/Others: $XDG_CONFIG_HOME/pyscaffold/default.cfg with fallback to ~/.config/pyscaffold/default.cfg

In order to provide these nice and compliant defaults, this commit
introduces a new vendorised package (`appdirs`) inside pyscaffold.contrib.
This is done to avoid adding dependencies to PyScaffold itself.
The `appdirs` module is very small and MIT-licensed, but rewriting its
features would be painful and require lots of efforts, specially for
testing. I assume that this feature is highly valuable for PyScaffold's
users (it gets really tedious after a while to type the same extensions
all the time), and therefore it is worthy.

There is an on going discussion about simplifying PyScaffold to not be
required in build time and therefore eliminating the needs for
vendorising dependencies. With the current written code, that is very
easy to change.

Future works not included in this PR:
- Add a `--config` flag to the CLI to specify passing a different config
  file/ `NO_CONFIG`.

Closes #236.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,7 @@ Version 4.0
 - Cookiecutter and Django extracted to their own repositories, issue #175
 - Support for Python 3.4 dropped, issue #226
 - Dropped deprecated ``requirements.txt`` file, issue #182
+- Added support for global configuration/PyScaffold's presets, issue #236
 
 
 Current versions
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,7 +11,8 @@ Version 4.0
 - Cookiecutter and Django extracted to their own repositories, issue #175
 - Support for Python 3.4 dropped, issue #226
 - Dropped deprecated ``requirements.txt`` file, issue #182
-- Added support for global configuration/PyScaffold's presets, issue #236
+- Added support for global configuration (avoid retyping common ``putup``'s
+  options), issue #236
 
 
 Current versions
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -148,6 +148,21 @@ venv. For example::
     .venv/bin/pip install tox
     .venv/bin/tox -e all
 
+..
+
+    I am trying to debug the automatic test suite, but it is very hard to
+    understand what is happening.
+
+`Pytest can drop you`_ in a interactive session in the case an error occurs.
+In order to do that you need to pass a ``--pdb`` option (for example by running
+``tox -- -k NAME_OF_THE_FALLING_TEST --pdb``).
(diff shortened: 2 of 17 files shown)


abravalheri:
@FlorianWilhelm the most controversial aspect of this feature is the vendorised `appdirs` (if we decide to not have PyScaffold at build time, that can be easily converted into dependency).

Finding a good way of referring to it in the docs was also tricky. My first reflex was to go inside `docs/configuration.rst` and write about it there... but then I realised that the term *configuration* was being used to describe the package's configuration not PyScaffold's, so that was confusing... For the time being I am calling it *preset*, but I am open to suggestions (also regarding in which page of the docs to put it).

FlorianWilhelm:
@abravalheri, I am fine with `appdirs` and hopefully when we moved to `pyproject.toml` we can just use it as a dependency.

abravalheri:
Thanks @FlorianWilhelm, I fixed the tests on windows, so I think I will merge this one since the feature was "pre-approved". We can do a big review before 4.x

abravalheri:
Thanks @FlorianWilhelm, I fixed the tests on windows, so I think I will merge this one since the feature was "pre-approved". We can do a big review before 4.x

PR #290, opened 2020-07-13
Remove dependency on PyScaffold during builds

### LINKED ISSUE #245: Remove pytest-runner setuptools integration
Recently pytest-dev/pytest#5546 was merged with a discussion about not recommending pytest-runner integration with setuptools.

Does this affect PyScaffold in any sense?

abravalheri:
As discussed in #256, after `pytest-runner` was deprecated and removed, the only strong reason for having PyScaffold required during builds is to use `setuptools_scm`. Docs can easily be generated via Makefile (already created by default) or `tox`/`nox` (the new `tox` template already contains a task for the docs).

A way of removing this dependency is by injecting `setuptools_scm` directly on the `setup.py` template. In the future something similar can be done for `pyproject.toml` (when it is generated by default).

This change is an attempt of doing that. As a direct consequence, we also remove PyScaffold as a build-time dependency of itself, which makes things easier. It also closes  #245.


abravalheri:
By the general comments in #256 and the quick hint in #289 (implying that in v4 PyScaffold will have runtime dependencies, but not "build-time" ones), I assume this PR is aligned with the current understanding on how PyScaffold should move forward. So now that the tests are fixed on Windows, I will merge the PR into v4.0.x.

The changes can also be reviewed/undone prior to the release.

Issue #309, opened 2020-08-03
pkg_resources "softly" deprecated.


abravalheri:
It seems that setuptools is committed to remove pkg_resources in the long run:

https://github.com/pypa/setuptools/milestone/4

Related: #218

PR #318, opened 2020-08-11
Remove pkg resources

abravalheri:
As stated in #309, `pkg_resources` is "kind of" deprecated, and on its way of being removed.

This PR is an attempt to replace it with other solutions

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -45,13 +45,13 @@ except FileNotFoundError:
 
 try:
     import sphinx
-    from pkg_resources import parse_version
 
     cmd_line_template = "sphinx-apidoc -f -o {outputdir} {moduledir}"
     cmd_line = cmd_line_template.format(outputdir=output_dir, moduledir=module_dir)
 
     args = cmd_line.split(" ")
-    if parse_version(sphinx.__version__) >= parse_version("1.7"):
+    if tuple(sphinx.__version__.split(".")) >= ("1", "7"):
+        # This is a rudimentary parse_version to avoid external dependencies
         args = args[1:]
 
     apidoc.main(args)
--- a/docs/features.rst
+++ b/docs/features.rst
@@ -72,12 +72,7 @@ Starting from Python 3.7 an even better approach is using `importlib.resources`_
 Note that we need a proper package structure in this case, i.e. directories need
 to contain ``__init__.py`` and we only specify the file ``data.txt``, no path is allowed.
 The library importlib_resources_ provides a backport of this feature.
-Even another way, provided by `setuptools`_'s  `pkg_resources`_ is::
 
-    from pkg_resources import resource_string
-    data = resource_string(__name__, 'path/to/my/data/relative/to/module.txt')
-
-Yes, actually "there should be one-- and preferably only one --obvious way to do it." ;-)
 
 .. _versioning:
 
@@ -346,7 +341,6 @@ Check out our :ref:`Configuration <default-cfg>` section to get started.
 .. _use TestPyPI: https://packaging.python.org/guides/using-testpypi/
 .. _importlib.resources: https://docs.python.org/3/library/importlib.html#module-importlib.resources
 .. _importlib_resources: https://importlib-resources.readthedocs.io/
-.. _pkg_resources: https://setuptools.readthedocs.io/en/latest/pkg_resources.html
 .. _flake8: http://flake8.readthedocs.org/
 .. _GitLab: https://gitlab.com/
 .. _PEP 440: http://www.python.org/dev/peps/pep-0440/
--- a/src/pyscaffold/templates/__init__.py
+++ b/src/pyscaffold/templates/__init__.py
@@ -4,7 +4,6 @@ Templates for all files of a project's scaffold
 
 import os
 import string
-from pkg_resources import resource_string
 from types import ModuleType
 from typing import Any, Dict, Set, Union
 
@@ -13,6 +12,18 @@ from configupdater import ConfigUpdater
 from .. import __version__ as pyscaffold_version
 from .. import dependencies as deps
 from .. import toml
+
+try:
+    from importlib.resources import read_text  # type: ignore
+except ImportError:
+    from pkgutil import get_data
+
+    def read_text(package, resource, encoding="utf-8") -> str:
+        data = get_data(package, resource)
+        if data is None:
+            raise FileNotFoundError(f"{resource!r} resource not found in {package!r}")
+        return data.decode(encoding)
+
 
 ScaffoldOpts = Dict[str, Any]
 
@@ -50,7 +61,7 @@ def get_template(
             is relative (in the standard module format, e.g. ``foo.bar.baz``).
             Notice that ``relative_to`` should not represent directly a shared
             namespace package, since this kind of package is spread in
-            different folders in the file sytem.
+            different folders in the file system.
 
             Default value: ``pyscaffold.templates``
             (**please assign accordingly when using in custom extensions**).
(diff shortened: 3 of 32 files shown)


docs/conf.py
```diff
@@ -45,13 +45,13 @@
...
 try:
     import sphinx
-    from pkg_resources import parse_version
 
     cmd_line_template = "sphinx-apidoc -f -o {outputdir} {moduledir}"
     cmd_line = cmd_line_template.format(outputdir=output_dir, moduledir=module_dir)
 
     args = cmd_line.split(" ")
-    if parse_version(sphinx.__version__) >= parse_version("1.7"):
+    if tuple(sphinx.__version__.split(".")) >= ("1", "7"):
```
abravalheri, line 53: Usually it is better to rely on proper parsing, but since Sphinx don't do something super fancy, I think we can trust this code (also passes on system tests)  and avoid dependencies.

docs/features.rst
```diff
@@ -72,12 +72,7 @@ Starting from Python 3.7 an even better approach is using `importlib.resources`_
 Note that we need a proper package structure in this case, i.e. directories need
 to contain ``__init__.py`` and we only specify the file ``data.txt``, no path is allowed.
 The library importlib_resources_ provides a backport of this feature.
-Even another way, provided by `setuptools`_'s  `pkg_resources`_ is::
 
-    from pkg_resources import resource_string
-    data = resource_string(__name__, 'path/to/my/data/relative/to/module.txt')
-
-Yes, actually "there should be one-- and preferably only one --obvious way to do it." ;-)
```
abravalheri, line 80: Sorry for removing the funny comment, I feel like since recommending just an old way and a new way of doing that, instead of 3 ways, it looses a little bit its impact.

src/pyscaffold/templates/__init__.py
```diff
@@ -14,6 +14,18 @@
...
+if sys.version_info[:2] >= (3, 7):
+    from importlib.resources import read_text
+else:
+    from pkgutil import get_data
+
+    def read_text(package, resource, encoding="utf-8") -> str:
+        data = get_data(package, resource)
+        if data is None:
+            raise FileNotFoundError(f"{resource!r} resource not found in {package!r}")
+        return data.decode(encoding)
```
abravalheri, line 26: Instead of doing that I could also add `importlib_resources` as a runtime dependency... but I felt like that was unnecessary for just one function that can be done with `pkgutil`.

src/pyscaffold/templates/__init__.template
```diff
@@ -1,10 +1,15 @@
-from pkg_resources import DistributionNotFound, get_distribution
+import sys
+
+if sys.version_info[:2] >= (3, 8):
+    from importlib.metadata import PackageNotFoundError, version
+else:
+    from importlib_metadata import PackageNotFoundError, version
```
abravalheri, line 6: This is the most controversial change in the entire PR, since it assumes we add `importlib_metadata` as a conditional runtime dependency (`install_requires`) in the generated package (which I include in the template and `update.py`)...

I feel like this would be more or less what the majority of people would do, and simply ditch `pkg_resources` completely (which is more future proof). Would it be OK @FlorianWilhelm?

The alternative is to keep using `pkg_resources` for Python < 3.8. Something like this:

```python
if sys.version_info[:2] >= (3, 8):
    from importlib.metadata import PackageNotFoundError, distribution
else:
    from pkg_resources import DistributionNotFound as PackageNotFoundError
    from pkg_resources import get_distribution as distribution

try:
    # Change here if project is renamed and does not equal the package name
    dist_name = ${distribution}
    __version__ = distribution(dist_name).version
except PackageNotFoundError:
    __version__ = "unknown"
finally:
    del distribution, PackageNotFoundError
```

src/pyscaffold/templates/setup_cfg.template
```diff
@@ -32,16 +32,16 @@ classifiers =
 
 [options]
 zip_safe = False
-packages = find:
+packages = find_namespace:
```
abravalheri, line 35: Alternatively we could choose dynamically between `find` and `find_namespace` depending on the `namespace` being activated or not...

Keeping everything with `find_namespace` is easier... and I don't think we have a drawback since we use the `src` layout (in the docs it is explicitly said `find_namespace_package` is backwards compatible with `find_package`).

src/pyscaffold/update.py
```diff
@@ -90,10 +98,43 @@ def add_entrypoints(struct: Structure, opts: ScaffoldOpts) -> "ActionParams":
...
+def add_dependencies(setupcfg: ConfigUpdater, opts: ScaffoldOpts):
+    """Add dependencies"""
+    options = setupcfg["options"]
+    if "install_requires" in options:
+        install_requires = options.get("install_requires", Object(value=""))
+        install_requires = deps.add(deps.RUNTIME, deps.split(install_requires.value))
+        options["install_requires"].set_values(install_requires)
+    else:
+        options.set("install_requires")
+        options["install_requires"].set_values(deps.RUNTIME)
```
abravalheri, line 129: Always add `import-metadata` backport as a dependency in the generated packages to allow fallback when finding the  version from setuptools_scm. This is controversial, see discussion in the [`__init__.template`](https://github.com/pyscaffold/pyscaffold/pull/318/files#r468882115).

docs/conf.py
FlorianWilhelm: Yes, agreed.

src/pyscaffold/templates/__init__.py
FlorianWilhelm: Yeah, makes sense.

docs/features.rst
FlorianWilhelm: Since it wasn't so funny, I guess it's okay to remove it ;-)

src/pyscaffold/templates/__init__.template
FlorianWilhelm: I like your PR solution. `importlib_metadata` is a backport of what's already established in 3.8 so you are exactly doing what the Python devs want. We could maybe add a `ToDo` comment stating that this can be removed when we have `python_requires = >=3.8` one day.

src/pyscaffold/templates/setup_cfg.template
FlorianWilhelm: If `find_namespace:` works I would also go for the simple solution since choosing dynamically just introduces another source of possible errors and needs more checks. I like simple solutions :-)

src/pyscaffold/update.py
FlorianWilhelm: Looks good for me. Also here a comment would be nice that this only needs to be done until `python_requires = >= 3.8`

FlorianWilhelm:
Looks good to me and can be merged. I would only suggest adding the comments as explained above.
Thanks a lot, @abravalheri for your work!

abravalheri:
Thank you very much @FlorianWilhelm for finding the time to review the PR! I know that you have been working very hard lately 😄. I am very happy to have your feedback!

I will add the comments and merge the PR. Also, I feel like this is the last major issue for PyScaffold 4 (at least in the core, not considering the extensions), so I will probably go ahead and release another v4 beta later this week (before that I might do some refactoring or reviewing/editing the docs/docstrings). If you have any other suggestion for changes in the API, specially if they break compatibility with v3, please let me know, maybe I can help.

src/pyscaffold/update.py
abravalheri: Question: are we adding a `python_requires = >= 3.6` by default in the generated `setup.cfg`?

FlorianWilhelm:
Thanks, @abravalheri! Regarding the API, I see no further necessary changes necessary except of handling the existing extensions. The `pyproject` extension would need to be deprecated. As far as I see we have in all extensions `setup.cfg`'s `install_requires = pyscaffold>=3.2a0,<4.0a0` so version 3.3 should be no problem (but tests needed) but all extensions for PyScaffold 4.0 would need to be updated.

Issue #235, opened 2019-07-14
On the road to PyScaffold 4.0


FlorianWilhelm:
## General discussion

This issue is a stub for everything that needs to be done and general discussion for the preparation of a PyScaffold 4.0 release which will include breaking changes.

FlorianWilhelm:
@abravalheri , let's discuss here what needs to be done to get the `v4.0.x` branch building again as well as the new extensions for django and cookiecutter. I think it makes sense to have `setup_requires` in a way that extensions either work with PyScaffold 3.2 or 4.0 to avoid version conflicts when we include breaking changes. Do you agree?

To fix the unit tests I already released an alpha version of PyScaffold 4.0 and I think I also need to do this for all the current extensions we have like dsproject, pyproject etc. I would do so by just choosing a higher version number and also releasing an alpha or beta version with corresponding `setup_requires` for PyScaffold 4.0. This would avoid the problems we are having right now with the PyScaffold 4.0 unit tests. Do you agree?

abravalheri:
Hi @FlorianWilhelm, great thinking!

Do you mean putting a cap on setup_requires (like you are doing nowadays for 3.2)? Something like: `setup_requires = pyscaffold>=4.0a0,<4.1a0`. I think this should be fine... Should we do `<4.1a0` or `<5.0a0`? Since are following semantic versioning, all the extensions that work on 4.1 will automatically work until 5.0. If we start with the cap in 5.0, when we release the next minor versions of PyScaffold, we can keep bumping the minimal. This way, we don't need to keep going to the extensions and update them.

Regarding the roadmap, I think we can address #215 and #219 relatively easily, because we have already a plan for that.

#218 is a little more tedious, since we have to go through all the extensions making sure the `__init__.py` files are empty and creating new releases.

#138 would be great to have. I have a proposal on how to rewrite `--update` as a built-in extension, but since what I envision it is back compatible, we can just go ahead with the initial plan.

#191 is more complicated though... In order to be able to have an interactive front-end (or something like an UI), we have to dissociate the concept of `parameters` and the internal implementation with `argparser`, and this might break the documented APIs. I will elaborate more in the open issue.

abravalheri:
As a new improvement, #236 would be excellent to have (however, since it is non-breaking improvement, it can be shipped in a minor version bump) 😄

FlorianWilhelm:
[...]

PR #324, opened 2020-08-21
Display a nice error when extension fails to load

abravalheri:
This is an attempt to solve the issue discussed in #235

### DIFF
--- a/src/pyscaffold/cli.py
+++ b/src/pyscaffold/cli.py
@@ -15,17 +15,11 @@ from .actions import ScaffoldOpts
 from .actions import discover as discover_actions
 from .dependencies import check_setuptools_version
 from .exceptions import exceptions2exit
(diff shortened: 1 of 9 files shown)


abravalheri:
Hi @FlorianWilhelm that definetely would be a very complete system test...
What I did instead was a more "unit-y" test (much easier to implement and run)... Maybe it is enough?

abravalheri:
I added another test that, while still being "unit-ish", is more realistic in the sense it actually attempts to load an extension from a python module, but that python module uses the old extension of the API (basically copied from the documentation in v3). I think this is a nice compromise between test simplicity + speed vs. realism...

Moreover I also tried to remove the multiple points of failure and the leaking abstraction by providing a common API for listing, loading and iterating extensions coming from entrypoints.

PR #331, opened 2020-11-10
Attempt to simplify complexity on #325

abravalheri:
As discussed in #325, trying to extend `argparse` is difficult and increases the complexity of the code, specially due to the "mixin" approach used in the original implementation.

This PR is an attempt to concretely show some of the ideas I mentioned in my review. 

(I suppose some of the errors in the CI will be solved once the interactive branch is rebased on v4.0.x)


FlorianWilhelm:
Really cool improvements! Thanks!

Issue #323, opened 2020-08-20
unexpected behavior with default coverage/pytest flags in setup.cfg


mathematicalmichael:
## Description of your problem

**Please provide a minimal, self-contained, and reproducible example.**
```python
putup project
```

The `setup.cfg` file will have a section like:

```
[tool:pytest]
# Options for py.test:
# Specify command line options as you would do when invoking py.test directly.
# e.g. --cov-report html (or xml) for html/xml output or --junitxml junit.xml
# in order to write a coverage file that can be read by Jenkins.
addopts =
    --cov project --cov-report term-missing
    --verbose
```

## Problem Behavior
Using interactive breakpoints in IDEs (vscode) is broken by the `--cov` related flags.
I _love_ that I can run `pytest` and `coverage report` and that's all it takes to generate a report. But is this going to definitely come at a cost of not having interactive debugging?

Please advise if there's a simple solution to this problem. I know it's a bit of a `setuptools` issue, but I was hoping there might be solution with the defaults chosen, or at the very least, a comment warning about this behavior, so users can comment out the flags during development but re-add them before pushing to CI pipelines.

## Versions and main components

* PyScaffold Version: 3.5.8
* Python Version: various. 3.7, 3.8
* Operating system: tested on Mac, Linux, CentOS container
* How did you install PyScaffold: pip

FlorianWilhelm:
Thanks for posting this issue. Can you please explain a bit more what you mean by broken? What happens exactly? Does the debugger hang or error?

I use the debugger via PyCharm quite often and never had a problem with it. I assume the problem only arises when you debug your unit tests, right?

mathematicalmichael:
https://stackoverflow.com/questions/40718760/unable-to-debug-in-pycharm-with-pytest seems to match it, so it looks like pycharm has settings that would override this behavior. perhaps vscode does too. 

And this problem was reported to me by someone using a setup.cfg I generated for a repo using pyscaffold, who works in the IDE, I believe he said the stack overflow solution with commenting out the flags is what worked for him, but perhaps there are settings in the application's debugger that can add overrides for the flags.

FlorianWilhelm:
Thanks for this pointer. We will add this to the docs as a warning starting from version 4.0, which is around the corner.

CarliJoy:
> while on that, I know it's unrelated, but a field / some documentation on how to add other pypi repositories for the install_requires would also be nice. I couldn't figure it out.

I don't know if that is actually possible.
[...]

PR #342, opened 2020-12-20
Add warning for users that --cov flags might cause trouble with breakpoints

FlorianWilhelm:
Fixing issue #323


FlorianWilhelm:
The failing checks are the same as fixed in PR #340. First merging #340 should fix them.

abravalheri (approved):

PR #348, opened 2020-12-22
Attempt to implement merge_user_input using parser

abravalheri:
Apparently not all validations from argparse happen when we call directly `Action.__call__`, as illustrated by some of the new tests that are failing.

Something that is also interesting to notice is that flag actions (`store_true` or `store_false`) seem to not be called at all when not present in argv, so calling `Action.__call__` with the other value does not yield the expected outcome.

### DIFF
--- a/docs/extensions.rst
+++ b/docs/extensions.rst
@@ -460,8 +460,11 @@ above, the automatically generated option will be ``--awesome-files``.
 
 For more sophisticated extensions which need to read and parse their
 own command line arguments it is necessary to override
-:obj:`~pyscaffold.extensions.Extension.augment_cli` that receives an
-:class:`argparse.ArgumentParser` argument. This object can then be modified
+:obj:`~pyscaffold.extensions.Extension.augment_cli` that receives as argument
+an object that implements a customised version of
+:obj:`~argparse.ArgumentParser.add_argument`.
+You can call :obj:`~pyscaffold.cli_parser.ArgumentParser.add_argument` on this
+object, as if you were working directly with :class:`argparse.ArgumentParser`
 in order to add custom command line arguments that will later be stored in the
 ``opts`` dictionary.
 Just remember the convention that after the command line arguments parsing,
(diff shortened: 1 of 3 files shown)


abravalheri:
@FlorianWilhelm this seem to work for some cases, but not every time (`store_true` or `store_false` seem to be particularly tricky)

FlorianWilhelm:
Hm... argparse is like the doorstep to hell. @abravalheri do you think we could use functionality from https://github.com/chriskiehl/Gooey ? I think what they accomplish is much more than what we want, they have a whole GUI but maybe we could learn from their concepts?

abravalheri:
Hi @FlorianWilhelm, thank you for the tip. After a quick look, it seems that Gooey build the entire CLI string and invoke the CLI command (?)

https://github.com/chriskiehl/Gooey/blob/2e8eaeefa29a280c725e66a54ca61e73534bd516/gooey/gui/containers/application.py#L92

We could create an list of each option and its values and pass it to the argparser...

FlorianWilhelm:
So what Gooey does is to extract a representation of argparse, do the GUI and sanity checks itself, create a string and feed it into argparse again. Sounds a bit complicated but could be a more promising approach, especially if their functionality could be reused? I don't know how stable their API is or if we would have to vendorize it.

abravalheri:
Yes. That might be more promising (although very overkill), I will send another commit here with another (failed) experiment I did... (fight against argparse is a nightmare).

abravalheri:
This last commit was just an attempt to create a `merge_user_input` inspired by the concept on Goey (just synthesise the complete CLI string).

The tests fail again since some actions do not accept all the available arguments for `add_argument` (or at least it seems so to me).

abravalheri:
Hi @FlorianWilhelm, I have the impression my last attempt might work. There is a bunch of logic branching, but that is more or less unavoidable with argparse...

Please feel free to review it and merge if you find relevant. Unfortunately there are not many automation regression tests that I could try to make sure it is not breaking anything...

CarliJoy:
Hey,

sorry got a bit lost in all the different places discussions happen :confused: 
I also look for existing projects and found:
https://github.com/TMiguelT/ArgparsePrompt

Some ideas:
 * Do not ask for every single input. Maybe we add an additional argument to add_argument `prompt_required=False`. Otherwise it is quite tiresome from user perspective
 * Lets use a only minimal modified ArgParse Class as base class that is extended only if "-i" is in the given arguments, so we won't brake things for the normal cli just to have the interactive mode.

abravalheri:
I am thinking of merging this to the `interactive` branch and them renaming that branch to `archives/interactive`...
This way we preserve the history of the previous interactive implementation in the case we ever need it in the future
(even if I merge the PR, the point prior to the merging will still be accessible via git history).

FlorianWilhelm:
Sounds like a good plan @abravalheri! This way we have it archived if we ever wanna go back on this track.

PR #357, opened 2020-12-28
[v4.0.x] Fix excluding options 

CarliJoy:
The options --config and --no-config should be exclusive.

For future interactive solutions this might be important, so this PR so early for so few changes.

Also note that if one is using tox -p more .coverage files are created (see git messages).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ tags
 # Unittest and coverage
 htmlcov/*
 .coverage
+.coverage.*.*
 .tox
 junit*.xml
 coverage.xml
--- a/src/pyscaffold/extensions/config.py
+++ b/src/pyscaffold/extensions/config.py
@@ -20,7 +20,8 @@ class Config(Extension):
     def augment_cli(self, parser: argparse.ArgumentParser):
         default_file = info.config_file(default=None)
         default_help = f" (defaults to: {default_file})" if default_file else ""
-        parser.add_argument(
+        group = parser.add_mutually_exclusive_group()
+        group.add_argument(
             "--config",
             dest="config_files",
             metavar="CONFIG_FILE",
@@ -28,7 +29,7 @@ class Config(Extension):
             type=Path,
             help=f"config file to read PyScaffold's preferences{default_help}",
         )
-        parser.add_argument(
+        group.add_argument(
             "--no-config",
             dest="config_files",
             action="store_const",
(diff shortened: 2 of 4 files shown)


abravalheri:
Thank you very much @CarliJoy for the PR.

I was trying to avoid using groups (specially when the user would not actively mix them, like in this case) -- when implementing the "interactive mode", we saw that having to override all the methods in argparse was really painful, so sticking with only `add_argument` would avoid that.

However, since in the last discussions (as in #333) it seems that we are more inclined to use the "editable mode" as an alternative to the interactive, maybe we don't have that problem anymore. I will try to see if the PR works with `--edit` and then I would be happy to merge it (sorry if it takes some time, I am trying to debug several things at the same time).

abravalheri:
Sorry the PR got automatically closed when I merged the v4 branch on master and deleted the branch, but it is still valid 😅

As I commented before I will try it against the "editable" branch and check if everything keeps working fine.

abravalheri:
Hi @CarliJoy thank you very much for the PR. I merged it manually into master (the only thing I left out was the gitignore thing, since I consider it a nice reminder to delete the coverage files if the tests are interrupted).

CarliJoy:
Hey @abravalheri 

thanks.
Only one confusion about the coverage files: They are already ignore when not run by parallel why ignoring them only if run in parallel?

CarliJoy:
But i guess this should / can be discussed in an extra PR ? Shall I open one?

abravalheri:
Hi @CarliJoy please feel free to open a new PR.  I am trying to understand your use case/tox workflow (so if you could expand a bit more the rationale behind the change, it would help a lot)...

If I am not wrong `tox -p` runs tox'es test environments in parallel. That was very useful in the past (for example for running `tox -p -e py27,py36`), however the default `tox.ini` file generated by PyScaffold was not really designed for it (in my opinion it makes little sense to run `tox -p -e py,clean,build,docs`). What PyScaffold adopts instead is [pytest-xdist](https://pypi.org/project/pytest-xdist/) to run the tests in parallel inside a single tox testenvironment (activated by `tox -- -n auto` [opt in feature, but that the user can easily change to be the default behaviour]).

With this option, the extra coverage files are still generated, but after pytest finishes running, the configuration that ships with PyScaffold make sure that they are combined in a single file and all the other ones are automatically deleted. They will not be deleted in the case the tox/pytest process is killed (e.g. CTRL+C).

I can see that sometimes someone might be surprised by a bunch of coverage files when they run `git status`, however in my mind they are a nice reminder that the tests did not finish running well and therefore you need to delete the files if you want to have your working directory clean. Ignoring the files will not make them go away... so if someone do a `ls` they would ending-up being surprised anyway...

PR #362, opened 2020-12-29
"Interactive mode" via editable file

abravalheri:
#333 got accidentally closed, this is a continuation of that PR.

TODO list:

- [x] Test more thoroughly and try to find/fix bugs
- [x] Hide private `argparse` functions (`parser._actions`, parser._actions) in a centralised location so they can be easily changed if necessary
- [x] Change short help text to something similar to: "Interactively choose and configure PyScaffold's parameters"
- [x] Rename `--edit` to `-i/--interactive`


docs/examples.rst
```diff
@@ -31,14 +43,29 @@ Just a few examples to get you an idea of how easy PyScaffold is to use:
...
+  This will create a project under the ``my_zope_subpackage`` directory with
+  the *installation name* of ``my-zope-subpackage`` (this is the name used by
+  pip_ and PyPI_), but with the following corresponding import statement::
+
+    from zope import subpackage
+    # zope is the namespace and subpackage is the package name
+
+  To be honest, there is really only the `Zope project <http://www.zope.org/>`_
   that comes to my mind which is using this exotic feature of Python's packaging system.
   Chances are high, that you will never ever need a namespace package in your life.
```
FlorianWilhelm, line 56: Great comment! I think statements like that help users a lot in figuring out which features are important for them.

src/pyscaffold/extensions/interactive.py
```diff
@@ -0,0 +1,282 @@
...
+    configurable = CONFIG.keys()
+    assert kind in configurable
+    initial_value = set(CONFIG[kind])
+    fallback_config: dict = {k: [] for k in configurable}
+
+    def _reducer(acc, ext):
+        config_from_ext = getattr(ext, CONFIG_KEY, fallback_config)
+        return acc | set(config_from_ext.get(kind, []))
+
+    return reduce(_reducer, list_all_extensions(), initial_value)
```
FlorianWilhelm, line 83: General note: It always takes me several minutes to understand what's happening in a `reduce` function and I think I am not the only one as Guido von Rossum once stated something similar. At that point, I have no better idea and this functional style is really smart but makes it also hard to understand what's going on.

src/pyscaffold/extensions/interactive.py
```diff
@@ -0,0 +1,282 @@
...
+    - ``"ignore"``: Options that should be simply ignored when creating examples
+    - ``"comment"``: Options that should be commented when creating examples,
+      even if they appear in the original :obj:`sys.argv`.
+
+Extension classes (or instances) can also provide configurations by defining a
+``interactive`` attribute assigned to a similar :obj:`dict` object.
+"""
+
+
+@lru_cache(maxsize=2)
```
FlorianWilhelm, line 65: Does this speed things up a lot? Just asking as introducing caching can always lead to quite subtle bugs later on.

src/pyscaffold/shell.py
```diff
@@ -8,12 +8,13 @@
...
 import sys
 from pathlib import Path
-from typing import Callable, Optional, Union
+from typing import Callable, Iterator, Optional, Union
 
 from .exceptions import ShellCommandException
 from .log import logger
 
 PathLike = Union[str, os.PathLike]
+EDITORS = ("sensible-editor", "nvim", "vim", "nano", "subl", "code", "notepad", "vi")
```
FlorianWilhelm, line 17: Should we have a comment here for the Sphinx module reference?

FlorianWilhelm (approved):
Really nice code! I really dig the implementation of the class Interactive. Tons of small functions are used like Lego to get the job done. That's clean code for me!

docs/examples.rst
abravalheri: Hi @FlorianWilhelm thanks for the comment, the original idea of mentioning Zope is yours 😝, I added the code example with the namespaces to make it more concrete...

src/pyscaffold/extensions/interactive.py
abravalheri: So the idea is that I call this function for every action, and everytime it loads the extensions from the entrypoints (so Python goes to setuptools files collecting entries and then do some just-in-time file loading). I am not sure how long it takes, I added the cache there just intuitively to be sincere. I agree the best would be to measure before doing any premature optimisation 😅 

To be fair, once all the extension classes and entrypoints are defined, this thing should be pretty much immutable (it is config metadata after all and works as a "distributed" constant across the extensions)... There could be problems if we are messing around with stubbing and mocking in the tests, but instead I am stubbing the `get_config` function itself instead of the things it depends on, so the cache is also taken away.

src/pyscaffold/shell.py
abravalheri: I will add that, thank you!

src/pyscaffold/extensions/interactive.py
abravalheri: I can revert it to a more procedural approach.

The main idea is to collect all the `interactive` configuration dicts from the extensions and merge them with `CONFIG`.

abravalheri:
Thank you very much for the review @FlorianWilhelm, I conducted a series of tests and tried to have 100% coverage on this extension... Hopefully it should work fine now. Just in case I added some docs saying it is experimental and asking for feedback from devs.

Hopefully this fulfils the interactive needs of PyScaffold.

docs/examples.rst
FlorianWilhelm: But the sentence `Chances are high, that you will never ever need a namespace package in your life.` is from you right?  That's the one I was referring to.

src/pyscaffold/extensions/interactive.py
FlorianWilhelm: Or you add your last sentence as a comment just before the return statement to make it easier for future contributors to understand what's going on.

FlorianWilhelm:
Yes! 🥳  This was the last open todo for version 4.0, right @abravalheri? As your PR resolves issue #191.

abravalheri:
Precisely! Now I guess it is just a matter of checking if the docs are in good shape, and we could have a release 😄 

I am happy in terms of features for v4. I think everything else can wait until v5 (even removing `--no-pyproject`), with the advantage that we can have some feedback.

FlorianWilhelm:
So far I found the docs in quite good shape. Regarding `--no-pyproject` we could leave it for the time being as you suggested. So whenever you feel like it, I am happy to see a release version 4.0 :-) The honor of releasing it should be all yours as you did like 99% of all the work. I cannot thank you enough for taking so much good care of PyScaffold.

abravalheri:
I will release a final beta before hand and wait a couple of weeks yet, just to be on the conservative side 😝 ...

Meanwhile I can have a look on the docs and also release updated versions of the extensions.

PR #369, opened 2021-01-02
Replace conda with vanilla python/tox in Github Actions

abravalheri:


.github/workflows/make-demo-repo.yml
```diff
@@ -20,17 +20,14 @@ jobs:
...
-        miniconda-version: "latest"
-        activate-environment: "pyscaffold"
-        environment-file: environment.yml
-        auto-update-conda: true
+        python-version: 3.7
         auto-activate-base: false
     - name: Build and create demo
       run: |
-        pip install -e . # otherwise templates are not packaged
+        python -m pip install -U setuptools pip -e .
```
FlorianWilhelm, line 30: Didn't know this is possible in one line without a semicolon. So the update and installation of `.` is done in one step? Cool.

.github/workflows/publish-package.yml
```diff
@@ -19,20 +19,14 @@ jobs:
...
-        environment-file: environment.yml
-        auto-update-conda: true
-        auto-activate-base: false
-    - name: Build and publish
+        python-version: 3.7
+    - name: Build Project and Publish
       env:
-        TWINE_USERNAME: "PyScaffold_Github"
+        TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
         TWINE_PASSWORD: ${{ secrets.PYPI_PYSCAFFOLD_GITHUB }}
```
FlorianWilhelm, line 29: Should we rather rename this to PYPI_PASSWORD? I mean otherwise, we are still giving away the username. In case we do, one would have to change that everywhere.

FlorianWilhelm (approved):
Looks good to me, just had a few comments.

.github/workflows/publish-package.yml
abravalheri: Hi @FlorianWilhelm, I think the renaming is a good idea. I can not do it, so I decided to leave it the way it was 😝 (Github does not allow renaming a secret, only creating a new one).

This template was copied from `--custom-extension`. I like the idea of having it as vars there at least, so if a third-party creates an the extension in they can simply configure a different user/password without changing the file.

.github/workflows/make-demo-repo.yml
abravalheri: I tested in the terminal a few times because I wanted the action to run as quick as possible (so less invocations of `pip`). It should work at least with a relatively recent version of `pip`.

.github/workflows/publish-package.yml
abravalheri: Also, the reason why I went for a `secrets.PYPI_USERNAME` instead of a `env.PYPI_USERNAME`, despite the username not being a secret, is that it seems to me that Github still does not allow organisation-wide env vars, just secrets, and configuring the username repo by repo would be annoying...

FlorianWilhelm: I created a secret `PYPI_PASSWORD` which is the same as `PYPI_PYSCAFFOLD_GITHUB` and removed the latter one. So we can change it in this PR to be more consistent. I will do it in the pyscaffoldext_dsproject extension.

abravalheri: Thank you very much. It should be fine in master now ;)

Issue #365, opened 2020-12-31
Should we use Github Actions to publish all the extension packages?


abravalheri:
Hi @FlorianWilhelm what do you think about using the same technique to publish all the extension packages (and maybe even configupdater) to PyPI? I suppose it will not consume loads of GA "build minutes" so it could be doable...

If we go for that, maybe `secrets.PYPI_PYSCAFFOLD_GITHUB` (or we create another account) could be shared as an [organisation secret](https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets#creating-encrypted-secrets-for-an-organization). so we can reuse it for all the projects?

FlorianWilhelm:
Hi @abravalheri, I think that's an awesome idea as I am growing more and more fond of Github actions :-) We could more or less copy the `publish-package` workflow scripts to any repository we want within the PyScaffold Organisiation. I moved the secret from the realm repository to organisation so that it is available for all repos. 

I also added the `API_GITHUB_TOKEN` to `pyscaffoldext-dsproject` so that I can also update the demo project `dsproject-demo` for it the same way we do it now for PyScaffold.

abravalheri:
Thank you very much @FlorianWilhelm, I added the `PyScaffold_Github` user to the repositories I have access on PyPI. When I have some time I will copy the publish-package to the extensions (I suppose it will be very handy and make it easier to develop the extensions).

Should we add the template of this action to `pyscaffoldext-custom-extension`?
Another doubt: in the default tox.ini template, I have `tox -e clean,build,publish` tasks that could be used in the action. Is it better to keep using `python setup.py` or use those tox tasks (which in turn rely on the new `build` module from PyPA)?

Enjoy new years eve!

FlorianWilhelm:
Thanks @abravalheri, are there any repos for which you are lacking access? 

Good idea, we could add it to `pyscaffoldext-custom-extension`, especially since we are so far the most frequent users as it seems :-) 

[...]

PR #373, opened 2021-01-03
Adopt wheel by default

abravalheri:
… now that we don't need PyScaffold to be a "setup-time" dependency, we
don't need sdists.

This PR also remove the old pyscaffold keyword argument from the
entrypoints (left behind in previous changes).

See discussion in #365.


FlorianWilhelm (approved):
lgtm :-)

PR #374, opened 2021-01-03
Adopt no-guess-dev instead of post-release for setuptools_scm

abravalheri:
… this version scheme is closer to the one we were using in PyScaffold v3 (previously we were using the `.post0.devX`, `"no-guess-dev"` used `.post1.devX`, while the current master branch uses `.postX`).
It requires `setuptools_scm > 5` (since it is a new implementation).

Additionally this PR updates the `setuptools` dependency to match the one from `pyproject.toml` (always a good thing to ensure the user will have a version of `setuptools` that supports PEP 517, 518 and 420).


FlorianWilhelm (approved):
LGTM. Since versioning is quite a crucial thing, I guess one should mention this small change in the CHANGELOG. Saying that semantically all stays the same but now all non-tag commits are versioned `LAST_TAG.post1.devN` instead of `LAST_TAG.post0.devN`.

Issue #378, opened 2021-01-07
No detailed error log during ErrorLoadingExtension


FlorianWilhelm:
## Description of your problem

Just produce e.g. an import error in an extension. When running `putup --dsproject ...` the error will be:
```
ERROR: There was an error loading 'pyscaffoldext-dsproject'.
    Please make sure you have installed a version of the extension that is compatible
    with PyScaffold 4.0b3.post1.dev1+ge4a1a38. You can also try unininstalling it.
```

In order to get to the actual cause of the error `putup -vv --dsproject ...` should work but due to the fact that `ErrorLoadingExtension` is a `RuntimeError` and the fact that the plugins are loaded before the error level is set, thus
```
  if logger.level <= logging.DEBUG:
      # user surely wants to see the stacktrace
      traceback.print_exc()
  print(f"ERROR: {ex}")
```
in `exceptions2exit` does not work as intended.

To solve this one could replace the condition `logger.level <= logging.DEBUG` and do some partial parsing of `sys.argv` with https://docs.python.org/3/library/argparse.html#partial-parsing. 

## Versions and main components

* PyScaffold Version: current master
* Python Version: 3.8
* Operating system: MacOS
* How did you install PyScaffold: pip

abravalheri:
Hi @FlorianWilhelm thank you for reporting this bug. It looks like a tricky one... I will try to have a look at it during the weekend (no promises :P).

Meanwhile, I think this might be solved by https://github.com/pyscaffold/pyscaffoldext-dsproject/pull/10 ...

FlorianWilhelm:
This was fixed by @abravalheri. Thank you!

PR #381, opened 2021-01-08
Use partial parsing to handle log_level before extensions

abravalheri:
This PR is an attempt to solve #378

### DIFF
--- a/src/pyscaffold/cli.py
+++ b/src/pyscaffold/cli.py
@@ -194,7 +194,7 @@ def parse_args(args: List[str]) -> ScaffoldOpts:
     add_extension_args(parser)
 
     # -- Parse options and transform argparse Namespace object into common dict
-    # Please not that there are many places you can process scaffold options.
+    # Please not that there are many places where you can process scaffold options.
     # We should only do the absolutely minimum necessary in the CLI-layer.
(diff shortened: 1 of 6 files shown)


FlorianWilhelm (commented):
I like the modularization by having a function `add_log_related_args`. What I am not 100% sure of is the added complexity by first parsing for the log_level and then parsing later again.

I am not sure if it would work but my approach would have been:
* also have a `add_log_related_args` function
* add a function `get_log_level()` that creates a new `parser = Argumentparser()` calls `add_log_related_args(parser)` and then `parser.parse_known_args` on `sys.argv` checking for the log_level and returning it. To my understandng the content of `sys.argv` is never changed and will always have the original arguments.
* In `exceptions2exit_decorator` one could then `if logger.level <= logging.DEBUG` replace with `if get_log_level() <= logging.DEBUG`

Do you think that would work too? And is maybe simpler?

abravalheri:
Hi @FlorianWilhelm thank you very much for the review.
Indeed the implementation is very complicated, and I was getting very puzzled, specially because `caplog` get a bit wild in the CI, and it is not making sense.

I will try your suggestions in a separated PR, but I will keep this one open for the time being until the other one is ready.

Do you mind if I ask your review again?

FlorianWilhelm:
Sure! Thanks for fixing this issue.

PR #393, opened 2021-01-24
Fix generated __init__.py no cover pragma

satwell:
The no cover pragma should apply to the entire except clause instead of
the statement in the clause.  Otherwise we get a coverage miss on the
"except" line.


abravalheri:
Thank you very much for the PR @satwell!

Issue #397, opened 2021-02-03
Module instead of package


ghuname:
I have just created a new project by using `putup scaffmodule` on my windows 10 machine.
I opened the project in pycharm, added file `/scaffmodule/src/scaffmodule/scaffmodule.py` with content:

```
def print_something():
    print("today is a lovely day")
```

![image](https://user-images.githubusercontent.com/15207713/106764735-5d08ed80-6638-11eb-8344-a2e3780e69a6.png)

Now I would like to create sdist. After executing `python setup.py sdist` from project root in dist folder a `scaffmodule-0.0.post0.dev1+gcb75cbb.dirty.tar.gz` file is created.

On another machine (centos 7) I have installed this sdist:

```
 pip install /root/scaffmodule-0.0.post0.dev1+gcb75cbb.dirty.tar.gz
Processing ./scaffmodule-0.0.post0.dev1+gcb75cbb.dirty.tar.gz
Building wheels for collected packages: scaffmodule
  Building wheel for scaffmodule (setup.py) ... done
  Created wheel for scaffmodule: filename=scaffmodule-0.0.post0.dev1+gcb75cbb.dirty-py2.py3-none-any.whl size=4786 sha256=5915a8978a6ac97232a7c3c02035d2208d699f6f09e7e0a76c8c2c7acf494989
  Stored in directory: /root/.cache/pip/wheels/24/be/da/2ef67b290c283ab504c8386c72eb959769a1c111699d194661
Successfully built scaffmodule
Installing collected packages: scaffmodule
Successfully installed scaffmodule-0.0.post0.dev1+gcb75cbb.dirty
```

Now when I try to use pyscaffoldmodule on this machine (centos 7) it looks like this:

```
>>> import scaffmodule
>>> dir(scaffmodule)
['__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__', '__version__', 'dist_name']
>>> import scaffmodule.scaffmodule
>>> dir(scaffmodule.scaffmodule)
['__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__', 'print_something']
>>> scaffmodule.scaffmodule.print_something()
today is a lovely day
```

What should I do in order to instead of `import scaffmodule.scaffmodule` use `import scaffmodule`, and consequentially instead of `scaffmodule.scaffmodule.print_something()` to use `scaffmodule.print_something()`.

[...]

Issue #162, opened 2018-05-16
__author__ / __copyright__ / __license__ not widely accepted as modern best practice


aspiers:
I was a bit surprised by the presence of `__author__`, `__copyright__`, and `__license__` being introduced to my project by PyScaffold, so I did a quick bit of googling to try to find whether there was any generally accepted wisdom about whether they should or shouldn't be used.  Most of the results seemed to reference pretty old style guides, e.g.

- http://web.archive.org/web/20111010053227/http://jaynes.colorado.edu/PythonGuidelines.html#module_formatting
- http://epydoc.sourceforge.net/manual-fields.html#module-metadata-variables (does *anyone* use epydoc these days?)

 I couldn't find *any* recent documents advising the use of these.  In contrast, look at the comments below [this answer to "What is the common header format of Python files?" on Stack Overflow](https://stackoverflow.com/questions/1523427/what-is-the-common-header-format-of-python-files/1523456#1523456).  Whilst the answer itself got a lot of votes simply by being helpfully factual without stating any opinion on best practice, by far the most upvoted comment was:

> I think all of this metadata after the imports is a bad idea. The parts of this metadata that apply to a single file (eg author, date) are already tracked by source control. Putting an erroneous & out of date copy of the same info in the file itself seems wrong to me. The parts that apply to the whole project (eg licence, versioning) seem better located at a project level, in a file of their own, rather than in every source code file.

And indeed with well-established modern best practices which involve [putting a single license file in the repository at the top level of the repository](https://help.github.com/articles/licensing-a-repository/#determining-the-location-of-your-license), and tracking authorship via git, these variables seem pretty anachronistic, in addition to violating the widely agreed-on [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).  So I suggest removing them.

FlorianWilhelm:
Thanks @aspiers for your research into this topic. The goal of PyScaffold is to define a standard for Python package creation by picking up well established community standards, making them easily accessible and drive new conventions like the `src`-layout for example. 

I agree to most of your points and I am actually not sure where I picked `__author__`, `__copyright__`, and `__license__` up. It might be that PyCharm was suggesting this for a while. 

That being said, when you create a new scaffold only the file `skeleton.py` will have those variables set. We can of course remove it from the template. Regarding the source code of PyScaffold itself I think we could also just drop it everywhere except in the `contrib` subpackages. @abravalheri: What do you think about this?

[...]

PR #398, opened 2021-02-04
Merge main and run in a single function inside skeleton

abravalheri:
As I went through #397 I found it difficult to explain/understand why we
have 2 functions in skeleton doing the CLI wrapping...

The docstring in `main` was also a bit confusing since it said that `main`
was the entry point for external calls. Since external calls can also
be interpreted as API calls from other packages, it sounded to me a bit weird that
people using skeleton's Python API should prefer passing stringfied
arguments instead of calling `fib` directly...

This change is a proposal to have `args` as an optional argument, which
makes it possible to unify the two functions into a single one, making
it a bit easier to explain.

I also added another test function for the CLI to the generated tests,
since we want to incentivise people to test.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -15,8 +15,8 @@ env:
 .regular_task_template: &REGULAR_TASK_TEMPLATE
   pip_cache: &pip-cache
     folder: $PIP_CACHE
-  # pre_commit_cache: &pre-commit-cache
-  #   folder: $PRE_COMMIT_HOME
+  pre_commit_cache: &pre-commit-cache
+    folder: $PRE_COMMIT_HOME
   tox_install_script:
     - python -m pip install --upgrade pip setuptools tox
   prepare_script: &prepare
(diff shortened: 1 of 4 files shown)


abravalheri:
```python
__author__ = "${author}"
__copyright__ = "${author}"
__license__ = "${license}"
```

I have a vague memory of one issue discussing how these lines are a bit of an old-fashion practice, given repositories have a `AUTHORS` and `LICENSE` files... More over it is not like a single file will ever be touched by a single author in the majority of the cases...

Should we remove them?

FlorianWilhelm:
> As I went through #397 I found it difficult to explain/understand why we
> have 2 functions in skeleton doing the CLI wrapping...

For me, it's `run` just for the `console_scripts` and `main` the actual function you implement. I agree that one could merge them into one but wouldn't that make the explanation even more complex? The current implementation I actually copied from somewhere else and I liked that the `main(args)` look a bit like C or Java and I hoped it would be familiar to many people. I agree that the comment is misleading about using this as an entry point for external calls. This we should remove. 

Regarding
```
__author__ = "${author}"
__copyright__ = "${author}"
__license__ = "${license}"
```
I have no strong feelings about this, so we could remove it. I copied the idea over from scipy/numpy and use it sometimes in customer projects. The point is that if you have a dedicated module (maybe some special algorithm or model code) then it's good to have like one or two persons that authored the code originally and are thus good people to talk to if you have general questions about the design and/or architecture.

abravalheri:
I found the original issue for the dunder variables (#162) and since no one have complained since then, I suppose we can leave as it is...

I changed the `run` and `main` functions back after your comment. Do you think the changes in documentation are worthy merging the PR?

Sorry if I hijacked the PR for investigating the mysterious cache problem with pyscaffoldext-dsproject... I can separate things in 2 PRs if you prefer.

FlorianWilhelm (approved):
Thanks a lot, @abravalheri. Looks good!

Issue #400, opened 2021-02-11
Add semver example in generated setup.cfg


CarliJoy:
## Describe your use-case
Currently I am always looking in the pyscaffold own setup.cfg to figure out how to provide semver like dependencies (which a best parctise).
Also I know that a lot of not so experienced dev aren't familar with it, so it would be good to nose point them to it ;-)

## Describe the solution you would like to see for your use-case
Simply convert in `src/pyscaffold/templates/setup_cfg.template`
```
# Add here dependencies of your project (line-separated)
# TODO: Remove conditional dependencies according to `python_requires` above
install_requires =
```
to 
```
# Add here dependencies of your project (line-separated)
# TODO: Remove conditional dependencies according to `python_requires` above
install_requires =
# Example:
# install_requires =
#    requests>=2.2,<3.0
# Note this makes sure to not run into api brakes,
# assuming the required packages follows Sementic versioning
# see https://semver.org/ for more information
```

PS: I don't understand the todo

FlorianWilhelm:
Thanks @CarliJoy, I think that's a good idea. Maybe we could change the text to something like
```
# install_requires =
#    requests>=2.2,<3.0
# Version specifiers like >=2.2,<3.0 avoid problems due the
# API changes in new majors versions. This works if the required
# packages follow Semantic versioning. For more information,
# check out https://semver.org/.
```

I am also not quite sure what
```
# TODO: Remove conditional dependencies according to `python_requires` above
```
means. Might be some reminiscent of version 4.0 changes, we should remove.

abravalheri:
Regarding the TODO comment, in [v4](https://github.com/pyscaffold/pyscaffold-demo/blob/63fe1cf4ee5cc969c34f80ea5d749ed83bb19c82/setup.cfg#L47) we have the following lines being generated:

```cfg
install_requires =
    importlib-metadata; python_version<"3.8"
```

If `python_requires = >=3.8`, then this dependency is never being activated and should be removed...

FlorianWilhelm:
@abravalheri: What if we would set 
```
python_requires = >=3.8
```
as a sane default in version 4.0 and have a comment:
```
# If your project requires or needs to support Python < 3.8, please add
#         importlib-metadata; python_version<"3.8"
# to `install_requires =` below.

abravalheri:
Thank you very much @FlorianWilhelm, I think setting `python_requires >= 3.8` by default is a bit complicated...
[...]

PR #402, opened 2021-02-12
Add comment about install_requires and semver

FlorianWilhelm:
This addresses issue #400. 

I just removed the `ToDo` part as we have this also in the file `dependencies.py`. This will remind us to drop 
```
importlib-metadata; python_version<"3.8"
```
when Python 3.7 is obsolete.


abravalheri (approved):
Thank you very much @FlorianWilhelm, it looks very good to me 😄 

In defence of my previous self that added the TODO comment I have to say that the comment in `setup.cfg` was there to remind the final users (since they actively might choose a different range for `python_requires` than the one we support) and the one in the `dependencies` for us the maintainers... But I agree that we don't have to take the users by the hand and prevent every single slightly not optimal configuration :P

PR #403, opened 2021-02-13
Remove 'all' extras from tox.ini

### LINKED ISSUE #401: tox.ini with undefined `all` in extras requirement...
By default in v4, the tox.ini template instructs tox to install `.[all]` and `.[testing]` dependencies, however we only define `testing` in `setup.cfg > extras_require not `all`...

https://github.com/pyscaffold/pyscaffold/blob/de27e2ed8cd8825282a7565f95513a39ec22f628/src/pyscaffold/templates/tox_ini.template#L10-L21

https://github.com/pyscaffold/pyscaffold/blob/de27e2ed8cd8825282a7565f95513a39ec22f628/src/pyscaffold/templates/setup_cfg.template#L54-L63

This results in a warning when running tox:
> **WARNING**:  `<package>` does not provide the extra 'all'

We should at least comment the `all` out or simply remove it...

abravalheri:
This change fixes #401.

### DIFF
--- a/src/pyscaffold/templates/tox_ini.template
+++ b/src/pyscaffold/templates/tox_ini.template
@@ -15,7 +15,6 @@ setenv =
 passenv =
     HOME
 extras =
-    all
     testing
 commands =
     pytest {posargs}


FlorianWilhelm:
Looks good! Thanks.

abravalheri:
Thank you very much for the review 😄

Issue #404, opened 2021-02-17
exclude doesn't exclude folders


Diyago:
## Description of your problem

**Please provide a minimal, self-contained, and reproducible example.**
I would like to exclude tests, images and Research folder from the list, but they anyway included to it. Full code here https://github.com/Diyago/GAN-for-tabular-data
_setup.cfg_
```
[options.packages.find]
where = src
exclude =
    tests
    Research
    images
```

**Please provide the full traceback.**
```python
no error
```

**Please provide any additional information below.**

## Versions and main components

* PyScaffold Version: 3.3
* Python Version: 3.8
* Operating system: Mac os
* How did you install PyScaffold: (conda/pip) pip

FlorianWilhelm:
Hi @Diyago, I just cloned your repo and ran:

> ❯ python setup.py bdist_wheel
running bdist_wheel
running build
running build_py
running egg_info
writing src/tabgan.egg-info/PKG-INFO
writing dependency_links to src/tabgan.egg-info/dependency_links.txt
writing requirements to src/tabgan.egg-info/requires.txt
writing top-level names to src/tabgan.egg-info/top_level.txt
writing manifest file 'src/tabgan.egg-info/SOURCES.txt'
installing to build/bdist.macosx-10.9-x86_64/wheel
running install
running install_lib
creating build/bdist.macosx-10.9-x86_64/wheel
creating build/bdist.macosx-10.9-x86_64/wheel/_ctgan
copying build/lib/_ctgan/models.py -> build/bdist.macosx-10.9-x86_64/wheel/_ctgan
copying build/lib/_ctgan/__init__.py -> build/bdist.macosx-10.9-x86_64/wheel/_ctgan
copying build/lib/_ctgan/README.MD -> build/bdist.macosx-10.9-x86_64/wheel/_ctgan
copying build/lib/_ctgan/transformer.py -> build/bdist.macosx-10.9-x86_64/wheel/_ctgan
copying build/lib/_ctgan/synthesizer.py -> build/bdist.macosx-10.9-x86_64/wheel/_ctgan
copying build/lib/_ctgan/conditional.py -> build/bdist.macosx-10.9-x86_64/wheel/_ctgan
copying build/lib/_ctgan/sampler.py -> build/bdist.macosx-10.9-x86_64/wheel/_ctgan
creating build/bdist.macosx-10.9-x86_64/wheel/tabgan
[...]

PR #406, opened 2021-02-23
Clearly state that wheels are the way to go now.

FlorianWilhelm:
Just as the title says and should help users with issues like in #404.


abravalheri (approved):

PR #410, opened 2021-02-28
Depend on ConfigUpdater 2.0

FlorianWilhelm:
## Purpose
Prepare PyScaffold for ConfigUpdater Version 2.0.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -40,7 +40,7 @@ package_dir =
 install_requires =
     importlib-metadata; python_version<"3.8"
     appdirs>=1.4.4,<2
-    configupdater>=1.1.3,<2
+    configupdater>=2.0,<3
     setuptools>=46.1.0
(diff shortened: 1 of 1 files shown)


abravalheri:
So it does not cause any compatibility problems or API changes? Fantastic...

Great job @FlorianWilhelm. How about we merge it immediately after v4.0 release and leave some beta or rc time for testing and feedbacks?

FlorianWilhelm:
Sounds really good! Let's do that.

abravalheri:
I merged the PR and also updated the version of pyscaffoldext-django to v0.1.1 because of a minor bug I found yesterday (that should not affect any new installation of v4.0, unless the person already have pyscaffoldext-django v0.1 installed).

Probably later today, after the tests run, I will tag v4.0.1rc1 so people can play around and report bugs.

PR #415, opened 2021-03-08
Add myself to contributors (for conda-forge work)

maresb:
## Purpose
@FlorianWilhelm [kindly suggested crediting me as a contributor](https://github.com/conda-forge/pyscaffold-feedstock/issues/29#issuecomment-792496301).

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -22,3 +22,4 @@ Contributors
 * Henning Häcker
 * Noah Pendleton
 * John Vandenberg
+* Ben Mares


FlorianWilhelm:
Thanks @maresb!

@abravalheri: Do you know why sometimes we see in the coverage test lately:
```
coveralls
Not on TravisCI. You have to provide either repo_token in .coveralls.yml or set the COVERALLS_REPO_TOKEN env var.
  File "/root/.local/lib/python3.6/site-packages/coveralls/cli.py", line 63, in main
    service_name=options['--service'])
  File "/root/.local/lib/python3.6/site-packages/coveralls/api.py", line 47, in __init__
    self.ensure_token()
  File "/root/.local/lib/python3.6/site-packages/coveralls/api.py", line 62, in ensure_token
    self.config_filename))
coveralls.exception.CoverallsException: Not on TravisCI. You have to provide either repo_token in .coveralls.yml or set the COVERALLS_REPO_TOKEN env var.
```
We don't use TravisCI anymore, right? Any idea how to fix this?

abravalheri:
I suppose coveralls do some magic to infer the tokens when running on travis, which is not the case any more...

The env var the error mention **is** configured in the cirrus.yml file:

```
COVERALLS_REPO_TOKEN: ENCRYPTED[...]
```
Maybe Cirrus CI started complaining when a non maintainer tries to decrypt the env var? (Sorry I added the encryption by following the docs of both Cirrus and Coveralls, not that I particularly believe it is fundamental). I triggered a new re-run... let's see if it works now.

Either way, this PR can be merged, if the coveralls keep complaining when it reaches the master, I will open a new issue to investigate it better.

abravalheri:
As I suspected, after merged, the CI does not run into the same error...

FlorianWilhelm:
Fascinating and thanks for giving this explanation. I would have totally gone in the wrong direction. Is there a solution for this? Imagine someone posts a feature PR and we cannot see if the feature is tested well enough due to this problem. Also, it might confuse new contributors.

abravalheri:
I am not exactly sure to be sincere...

I added encryption on the commit https://github.com/pyscaffold/pyscaffold/commit/20e45905, because the documentation suggests that, if the token is present it should be encrypted.
But the token itself was introduced on the commit: https://github.com/pyscaffold/pyscaffold/commit/9ef6775af6e13b434bf9adfdc321a15974487eaa, supposedly because the tests were breaking when the token was not available (as in here https://github.com/pyscaffold/pyscaffold/runs/231087719).

Coveralls do not support cirrus-ci by default, but the docs suggest it can be easily done by setting [some environment variables](https://docs.coveralls.io/supported-ci-services):
```
CI_NAME
CI_BUILD_NUMBER
CI_BUILD_URL
CI_BRANCH
CI_PULL_REQUEST (optional)
```

Maybe if these variables are available the token could be omitted (just guessing here).
One thing left to check is if Cirrus CI can be configured to allow decoding of secrets on PR (which I find more difficult).

Issue #418, opened 2021-03-09
Setuptools is deprecating dash-separated keys on setup.cfg


abravalheri:
It seems that the latest versions of `setuptools` are deprecating dash-separated keys on `setup.cfg`, and we have a mix of them our template for `setup.cfg`.

The warnings can be seen by running `tox -e build`
```
/tmp/build-env-cjhu26rf/lib/python3.8/site-packages/setuptools/dist.py:634: UserWarning: Usage of dash-separated 'author-email' will not be supported in future versions. Please use the underscore name 'author_email' instead
  warnings.warn(
/tmp/build-env-cjhu26rf/lib/python3.8/site-packages/setuptools/dist.py:634: UserWarning: Usage of dash-separated 'project-urls' will not be supported in future versions. Please use the underscore name 'project_urls' instead
  warnings.warn(
/tmp/build-env-cjhu26rf/lib/python3.8/site-packages/setuptools/dist.py:634: UserWarning: Usage of dash-separated 'long-description' will not be supported in future versions. Please use the underscore name 'long_description' instead
  warnings.warn(
/tmp/build-env-cjhu26rf/lib/python3.8/site-packages/setuptools/dist.py:634: UserWarning: Usage of dash-separated 'long-description-content-type' will not be supported in future versions. Please use the underscore name 'long_description_content_type' instead
  warnings.warn(
/tmp/build-env-cjhu26rf/lib/python3.8/site-packages/setuptools/dist.py:634: UserWarning: Usage of dash-separated 'no-vcs' will not be supported in future versions. Please use the underscore name 'no_vcs' instead
  warnings.warn(
/tmp/build-env-cjhu26rf/lib/python3.8/site-packages/setuptools/dist.py:634: UserWarning: Usage of dash-separated 'max-line-length' will not be supported in future versions. Please use the underscore name 'max_line_length' instead
  warnings.warn(
/tmp/build-env-cjhu26rf/lib/python3.8/site-packages/setuptools/dist.py:634: UserWarning: Usage of dash-separated 'extend-ignore' will not be supported in future versions. Please use the underscore name 'extend_ignore' instead
  warnings.warn(
```

abravalheri:
@FlorianWilhelm do you think we should do automatic conversions inside `update.py`, or just change the templates? (the first option is more cumbersome to implement, but it would fix problems transparently to the users; the second option is very lightweight and could even be considered in a bug-fix release)

abravalheri:
[...]

PR #419, opened 2021-03-09
Replace '-' with '_' in setup.cfg

abravalheri:
This PR addresses #418
(it should fix the issue for new projects, but not affect existing ones).

### DIFF
--- a/docs/example_setup.cfg
+++ b/docs/example_setup.cfg
@@ -5,10 +5,10 @@
 name = my_project
 description = A test project that was set up with PyScaffold
 author = Florian Wilhelm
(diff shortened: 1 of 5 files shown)


FlorianWilhelm:
Thanks, @abravalheri! Fixing the template seems like a really good and lightweight solution to me. Users of existing projects would need to run `putup -u -f` to fix this, right?

abravalheri:
Yes, I believe doing a forced update and merging the changes back would suffice (unless there is something really obscure going on :P)

abravalheri:
Now that the tests pass, I will merge this PR... Updating existing code bases is something additional to changing the templates.

Issue #135, opened 2017-12-05
Anaconda support


FlorianWilhelm:
Since more and more people use conda packages, a flag `--conda` should create all files necessary to build Anaconda packages.

FlorianWilhelm:
PyScaffold already has an conda package at [conda-forge](https://github.com/conda-forge/pyscaffold-feedstock).

FlorianWilhelm:
Maybe it's best to support this first with the help of an official PyScaffold extension package. I'll drop this as an requirement for core PyScaffold 3.1. Another reasons is that actually no PyScaffold user ever requested this feature nor commented on this ticket.

willu47:
I think this would be very useful. I'm using Pyscaffold for the packages [SALib](/SALib/SALib) and [smif](/nismod/smif), and having trouble with the conda-forge feedstocks for [SALib](/conda-forge/salib-feedstock).

FlorianWilhelm:
I am not sure this issue makes any more sense. The problem is that in order to create a conda package, e.g. for [conda-forge](https://conda-forge.org/), you have an extra repository with a kind of build recipe, aka *feedstock*. [Here is the Python Feedstock as an example](https://github.com/conda-forge/pyscaffold-feedstock). This also means that it makes no sense to create this structure within the current template as it would rather be completely independent in its own repo. Consequently, a `--conda` flag makes no sense. To create a conda feedstock, one could just mention in the docs that everything is explained under https://conda-forge.org/docs/maintainer/adding_pkgs.html.

abravalheri:
Let me see if I got it right:

Anaconda allows users to build and upload packages and this process is described in [conda-build/build-pkgs](https://docs.conda.io/projects/conda-build/en/latest/user-guide/tutorials/build-pkgs.html) and [conda-build/build-pkgs-skeleton](https://docs.conda.io/projects/conda-build/en/latest/user-guide/tutorials/build-pkgs-skeleton.html). These recipes can in theory coexist within the same repository as generated by PyScaffold.

The problem with this approach is that Anaconda is organised in channels, and users can only upload to their own channels, but mixing packages from different channels is generally not a good idea, because there might be clashes and dependency hell...

There are separated communities such as conda-forge (the most prominent one for Python and in special for data science) and bioconda, but these communities have their own procedures for publishing which can be orthogonal with the idea of package code and recipe code coexisting inside the same repository.

[...]

PR #424, opened 2021-03-11
Add conda to docs

abravalheri:
This PR is related to the discussion on #135


FlorianWilhelm (approved):
Really good introduction I think.

abravalheri:
First I thought about adding some comment on the FAQ (which is usually what I do when I don't have time/brain to compose a proper text explaining things), but then it kind of evolved... Moreover `conda` is important enough to have its own section...

The rendered version can be seen at [readthedocs](https://pyscaffold.org/en/latest/conda.html)

Thank you for the review @FlorianWilhelm

maresb:
This is a really nice writeup. I'm very impressed.

A potential problem I see is that "Creating Conda Packages" is a very niche topic which 99% of users will probably never do. But using `conda` in conjunction with PyScaffold in particular `dsproject` is a very important mainstream theme. This niche topic is now very prominent in the PyScaffold documentation, while (as far as I can tell) there isn't really a prominent explanation of using `conda` in general. I think this gives the wrong impression about what `conda` is mainly for. I wonder if it would be better to: 

* rename the topic "Creating Conda Packages" → "Using Conda Packages",
* describe using `conda` as a package manager, the `environment.yml` file, etc, and
* put this section which you just wrote at the end.

abravalheri:
Thank you very much for the feedback @maresb, I really like the suggestions.
I will open a new issue so we can track the progress in this area.

Hopefully we can re-use some of Florian's previous explanations in #368 and #426 to compose the extra sections.

Issue #428, opened 2021-03-17
Quickstart Fails: cannot import name 'VERSION_CLASS'


DMPS:
Hi Pyscaffold team,

I am trying to bootstrap a new project with PyScaffold using the quickstart here: [Pypi](https://pypi.org/project/PyScaffold/#description). PyScaffold 4.0 installs perfectly fine, but I get the following error when I run `putup my_project`:

```bash
Traceback (most recent call last):
  File "/Users/daltonscott/Library/Python/3.7/bin/putup", line 5, in <module>
    from pyscaffold.cli import run
  File "/Users/daltonscott/Library/Python/3.7/lib/python/site-packages/pyscaffold/cli.py", line 13, in <module>
    from . import api, templates
  File "/Users/daltonscott/Library/Python/3.7/lib/python/site-packages/pyscaffold/api.py", line 9, in <module>
    from . import actions, info
  File "/Users/daltonscott/Library/Python/3.7/lib/python/site-packages/pyscaffold/actions.py", line 20, in <module>
    from . import info, repo
  File "/Users/daltonscott/Library/Python/3.7/lib/python/site-packages/pyscaffold/info.py", line 30, in <module>
    from .templates import ScaffoldOpts, licenses, parse_extensions
  File "/Users/daltonscott/Library/Python/3.7/lib/python/site-packages/pyscaffold/templates/__init__.py", line 14, in <module>
    from .. import dependencies as deps
  File "/Users/daltonscott/Library/Python/3.7/lib/python/site-packages/pyscaffold/dependencies.py", line 9, in <module>
    from setuptools_scm.version import VERSION_CLASS
ImportError: cannot import name 'VERSION_CLASS' from 'setuptools_scm.version' (/Users/daltonscott/Library/Python/3.7/lib/python/site-packages/setuptools_scm/version.py)
```

What I find most odd, is that I setup a pyscaffold 4.0 project literally last Friday and I did not get any error. Has there been some change in the code since then?

abravalheri:
Hi @DMPS, this happened because the latest update on [`setuptools_scm`](https://github.com/pypa/setuptools_scm/commit/b059a6bc44e9dd5703ffa4cc3c116beacb7edfa6) (they just released a new version [5h ago](https://pypi.org/project/setuptools-scm/#history)).

While we work on the fix, could you please specifiy `setuptools_scm==5.0.2` in your installation for the time being?

abravalheri:
[...]

PR #429, opened 2021-03-17
Remove checks for old setuptools installation

abravalheri:
As discussed in #428, some checks we were making for the setuptools
version are incompatible with the latest version of setuptools_scm.

So one alternative (implemented in this PR) is to simply stop verifying
the versions and assume they are satisfied because they are correctly
declared as `install_requires` in `setup.cfg`.

The only doubt left is: what happens if the person is building a
generated project with `--no-pyproject`?

(In `pyproject.toml` we do have a minimum requirement for setuptools,
which is more than enough, but in `setup.py` that is omitted... Should we
add one? Is that a good practice? Or are we just being optimistic and
relying on the error message asking the user to update?)

### DIFF
--- a/tests/test_install.py
+++ b/tests/test_install.py
@@ -135,7 +135,8 @@ class DemoApp(object):
 
     def setup_py(self, *args, **kwargs):
         with chdir(self.pkg_path):
-            args = ["python", "setup.py"] + list(args)
+            args = ["python", "-Wignore", "setup.py"] + list(args)
+            # Avoid warnings since we are going to compare outputs
             return self.run(*args, **kwargs)
 
     def build(self, dist="bdist", cli_opts=()):
--- a/setup.cfg
+++ b/setup.cfg
@@ -45,7 +45,14 @@ install_requires =
     setuptools_scm>=5
     tomlkit>=0.7.0,<2
     packaging>=20.7
+
 # packaging is versioned by year, not SemVer
+
+# Notes about setuptools versions:
+# - 40.1: required for `find_namespace`
+# - 45: required for `setuptools_scm` v6
+# However we specify a higher version so we encourage users to update the
+# version they have installed...
 
 [options.packages.find]
 where = src
--- a/src/pyscaffold/cli.py
+++ b/src/pyscaffold/cli.py
@@ -13,7 +13,6 @@ from . import __version__ as pyscaffold_version
 from . import api, templates
 from .actions import ScaffoldOpts
 from .actions import discover as discover_actions
-from .dependencies import check_setuptools_version
 from .exceptions import exceptions2exit
 from .extensions import list_from_entry_points as list_all_extensions
 from .identification import get_id
@@ -253,7 +252,6 @@ def main(args: List[str]):
     Args:
         args: command line arguments
     """
-    check_setuptools_version()
     opts = parse_args(args)
     opts["command"](opts)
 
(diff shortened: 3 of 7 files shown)


abravalheri:
No dev can be truly happy without windows errors...

FlorianWilhelm:
Thanks, @abravalheri, I think this is a simple and good solution. We depend on the right version of `setuptools_scm` and let it do its checks. 

Regarding the `--no-pyproject` flag (and here it bites us already ;-). I see two options:
* Remove the flag in `4.0.1` and call it a bug that it existed otherwise we would need `5.0`.
* Also make sure that it is specified in `setup.cfg` somehow.

BTW, how does it work right now if someone uses `--no-pyproject`? Where is `setuptools-scm` as development requirement defined? `setup_requires` doesn't exist anymore, right?

tests/test_install.py
```diff
@@ -135,7 +135,8 @@ def cli(self, *args, **kwargs):
 
     def setup_py(self, *args, **kwargs):
         with chdir(self.pkg_path):
-            args = ["python", "setup.py"] + list(args)
+            args = ["python", "-Wignore", "setup.py"] + list(args)
```
abravalheri, line 138: I suspect something like the following warning is happening in windows and causing the [error in the CI](https://cirrus-ci.com/task/5380876931432448?command=test#L672):
```
c:\python38\lib\site-packages\setuptools\distutils_patch.py:25: UserWarning: Distutils was imported before Setuptools. This usage is discouraged and may exhibit undesirable behaviors or errors. Please use Setuptools' objects directly or at least import Setuptools first.
  warnings.warn(
```
But I could not replicate it myself...
So I hope ignoring the error might fix the assertion 🤞

abravalheri:
Thanks for the review @FlorianWilhelm.

Right now there is no specification about build-time dependencies in `setup.cfg`. If I remember it well, there was a problem specifying the requirement versions and I opened a ticket in `setuptools` where they told me I should not be using `setup_requires`.

Instead we are relying on the logic implemented in `setup.py`, where we clear state that the user should install an updated version of `setuptools`, `setuptools_scm` and `wheel`.

There is a mention it in the update docs for people that don't want to use `pyproject.toml`:
```
The use of setup_requires is discouraged. When updating to v4 PyScaffold will remove this field automatically and transfer the dependencies to the pyproject.toml :: build-system.requires field, which means you may need to manually place them back when deleting pyproject.toml. Alternatively you can ditch setup_requires completely and rely on other tools like tox or make to build your project with the correct dependencies in place inside a virtual environment. This have the advantage of increasing reproducibility. With tox you can specify a build testenv with the skip_install option and the required build time dependencies listed in deps.
```
I wonder if this note should be place in a different place.

Although I am not against removing `--no-pyproject` right now (it will definitely be removed in v5), I suppose we could rely on the warning message inside `setup.py` to avoid breaking changes in a minor/patch update. If we do want to remove it now, I would keep the flag but print a help text instead.

FlorianWilhelm (approved):
Looks really good, thanks.

abravalheri:
@FlorianWilhelm, I am thinking about finally releasing 4.0.1.
I will have a look on the extensions to see if they are relying of setuptools dashed-keys. Is there anything else we should check?

I always feel a bit nervous when releasing new versions, afraid that something is missing.

FlorianWilhelm:
@abravalheri, thank you so much for taking care of this and so fast. 

Besides the extensions, I think that should be all for PyPI. Not sure what needs to be done for conda-forge.

abravalheri:
For conda-forge, hopefully the bot will prepare a pull request automatically for us. Otherwise we have to create a PR bumping the version and returning the `build` counter to 0.

PR #431, opened 2021-03-29
Quote shell.get_executable to avoid errors with whitespace

### LINKED ISSUE #430: Error when using interactive mode in windows: '...\Programs\Microsoft' is not recognized as an internal or external command, ...
It happens when i use the interactive mode. I got error message as below:

```bash
'C:\Users\***\AppData\Local\Programs\Microsoft' is not recognized as an internal or external command,
operable program or batch file.
ERROR:
```
(I omit my account info using ***.)
Same error happens on both powershell and cmd.
**Please provide a minimal, self-contained, and reproducible example.**
I try to locate the error, so i make a py script. It's part of the open file with editor. I got a familiar error.
```python
from pyscaffold.shell import *
from pyscaffold import file_system

with file_system.tmpfile(prefix="pyscaffold-", suffix=".args.sh") as file:
    content = edit(file).read_text("utf-8")
```
It seems the space in vscode path ruin the function `get_executable()` in `shell.py`.

**Please provide any additional information below.**
I'm not sure whether it's my computer's problem. But when i change **line 157** in `shell.py` from 
```python
        return executable
```
to 
```python
        return name
```
There is no error. 

And I wonder if the function `get_excutable()` could deal with any space in path of env, editor or any other commands.

* PyScaffold Version: 4.0.1
* Python Version: not specified
* Operating system: Win10
* How did you install PyScaffold: conda

abravalheri:
Attempt to fix #430, by `shlex.quote`-ing the executable path.

### DIFF
--- a/src/pyscaffold/shell.py
+++ b/src/pyscaffold/shell.py
@@ -159,34 +159,43 @@ def get_executable(
         prefix: look on this directory, exclusively or in additon to $PATH
             depending on the value of ``include_path``. Defaults to :obj:`sys.prefix`.
         include_path: when True the functions tries to look in the entire $PATH.
-        quote: quote return value if it contains whitespaces or other characters
-            considered special in a shell environment. ``True`` by default.
+
+
+    Note:
+        The return value might contain whitespaces. If this value is used in a shell
+        environment, it needs to be quote properly to avoid the underlying shell
+        interpreter splitting the executable path.
     """
     executable = shutil.which(name)
-    if not (include_path and executable):
-        candidates = list(Path(prefix).resolve().glob(f"*/{name}*"))
-        # ^  this works in virtual envs and both Windows and POSIX
-        if candidates:
-            path = (f.parent for f in sorted(candidates, key=lambda p: len(str(p))))
-            executable = shutil.which(name, path=os.pathsep.join(str(p) for p in path))
-            # ^  which will guarantee we find an executable and not only a regular file
+    if include_path and executable:
+        return executable
 
-    if quote and executable is not None:
-        return _quote(executable)
-        # ^  avoid problems with whitespace in executable path
+    candidates = list(Path(prefix).resolve().glob(f"*/{name}*"))
+    # ^  this works in virtual envs and both Windows and POSIX
+    if candidates:
+        path = (f.parent for f in sorted(candidates, key=lambda p: len(str(p))))
+        return shutil.which(name, path=os.pathsep.join(str(p) for p in path))
+        # ^  which will guarantee we find an executable and not only a regular file
 
-    return executable
+    return None
 
 
 def get_command(
-    name: str, prefix: PathLike = sys.prefix, include_path=True, **kwargs
+    name: str, prefix: PathLike = sys.prefix, include_path=True, shell=True, **kwargs
 ) -> Optional[ShellCommand]:
     """Similar to :obj:`get_executable` but return an instance of :obj:`ShellCommand`
     if it is there to be found.
     Additional kwargs will be passed to the :obj:`ShellCommand` constructor.
     """
-    executable = get_executable(name, prefix, include_path, quote=True)
-    return ShellCommand(executable, **kwargs) if executable else None
+    executable = get_executable(name, prefix, include_path)
+    if not executable:
+        return None
+
+    if shell:
+        executable = _quote(executable)
+
+    kwargs["shell"] = shell
+    return ShellCommand(executable, **kwargs)
 
 
 def get_editor(**kwargs):
@@ -203,7 +212,8 @@ def get_editor(**kwargs):
 
 def edit(file: PathLike, *args, **kwargs) -> Path:
     """Open a text editor and returns back a :obj:`Path` to file, after user editing."""
-    editor = ShellCommand(get_editor())
+    editor = ShellCommand(_quote(get_editor()))
+    # since we quote here we cannot pass flags in EDITOR environment variable
     editor(file, *args, **{"stdout": None, "stderr": None, **kwargs})
     # ^  stdout/stderr=None => required for a terminal editor to open properly
     return Path(file)
@@ -219,7 +229,7 @@ def _quote(executable_path: str) -> str:
     """Prevent whitespaces for breaking executable paths when called from
     shell.
 
-    Please not this is not intended to quote everything, only executable paths.
+    Please note this is not intended to quote everything, only executable paths.
     This assumption is important because we don't have to concern about special
     characters that are disallowed in file names. For example, on Windows
     ``"`` chars are reserved characters and not supposed to appear if file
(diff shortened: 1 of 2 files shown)


src/pyscaffold/shell.py
```diff
@@ -47,6 +51,10 @@ def run(self, *args, **kwargs) -> subprocess.CompletedProcess:
         """Execute command with the given arguments via :obj:`subprocess.run`."""
         params = subprocess.list2cmdline(list(map(str, args)))
         command = f"{self._command} {params}".strip()
```
abravalheri, line 53: One alternative not implemented in this PR is to pass `command` as a list of arguments to `subprocess.run`...
In that case Python should know how to handle whitespaces in the path...

I wonder how backward compatible that would be, and if it would break something in the main repository or in the extensions...

src/pyscaffold/shell.py
```diff
@@ -159,22 +175,29 @@ def get_executable(
...
     if it is there to be found.
     Additional kwargs will be passed to the :obj:`ShellCommand` constructor.
     """
     executable = get_executable(name, prefix, include_path)
-    return ShellCommand(executable, **kwargs) if executable else None
+    if not executable:
+        return None
+
+    if shell:
+        executable = _quote(executable)
```
abravalheri, line 197: Quoting here should be safe, since we are automatically obtaining the executable path, and checking for `shell`

src/pyscaffold/shell.py
```diff
@@ -191,7 +214,39 @@ def get_editor(**kwargs):
 
 def edit(file: PathLike, *args, **kwargs) -> Path:
     """Open a text editor and returns back a :obj:`Path` to file, after user editing."""
-    editor = ShellCommand(get_editor())
+    editor = ShellCommand(_quote(get_editor()))
+    # ^  since we quote here we cannot pass flags in EDITOR environment variable
```
abravalheri, line 218: I don't know how popular is to add flags to the `EDITOR` variable, but that is quite common with other variables.
For example, right now I have `PAGER=less -R` in my env.

This change prevents any attempt to add flags...

src/pyscaffold/shell.py
```diff
@@ -191,7 +214,39 @@ def get_editor(**kwargs):
...
+    ``"`` chars are reserved characters and not supposed to appear if file
+    paths (https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file).
+    """
+    if os.name == "posix":
+        return shlex.quote(executable_path)
+        # shlex is supposed to not quote if not necessary
+    elif os.name == "nt" and _has_whitespace(executable_path):  # pragma: no cover
+        # in our current CI setup, Windows-specific code might run during tests
+        # but the coverage data is not submitted
+        return f'"{executable_path}"'
```
abravalheri, line 249: This quote method should be fine, since quote symbols cannot be used in file paths on Windows... However quoting is always trick...

An alternative could be using Windows short path, as described in https://stackoverflow.com/questions/23598289/how-to-get-windows-short-file-name-in-python.

It would be a more convoluted way, but I suppose more robust.

src/pyscaffold/shell.py
FlorianWilhelm: I remember that I wrote `shell.py` back then only because I wanted to avoid dependencies. Is there maybe a package that does all this way easier?

src/pyscaffold/shell.py
FlorianWilhelm: I got a bad feeling with quoting here due to the implications you mentioned. One would trade the flexibility under Mac/Linux for fixing some Windows limitations.

src/pyscaffold/shell.py
FlorianWilhelm: Couldn't we just always quote the executable? Doesn't for instance `"ls"` work in unix environments?

src/pyscaffold/shell.py
abravalheri: One thing important to mention is that the problem with the spaces in the executable path can also be observed in POSIX environments 😝 

As long as we use `subprocess.run` with `shell=True`, we do need to escape the whitespaces, and therefore we prevent the ability to set flags via env vars.

One way around that, would be quoting the result in the `get_editor`, only when we know it does not come from environment variables, more specifically here:
https://github.com/pyscaffold/pyscaffold/blob/1f7f64f94b567e8bb6728d4cc5ad847f84366df9/src/pyscaffold/shell.py#L182

```python
# instead we could do
others = (_quote(get_executable(e)) for e in EDITORS)
```
and then require the users to do proper quoting if they rely on environment variables for `EDITOR` or `VISUAL`.

However we are coupling `get_editor` with a specific way of calling `ShellCommand`... If we are only interested in the raw value of the editor executable path (for example to compare with other path) then we end up with extra problematic quotes (e.g. if later we decide to call `subprocess.run` or even instantiate `ShellCommand` with `shell=False` and an arg list, we would have an error because quotes are not valid chars in an executable path).

src/pyscaffold/shell.py
abravalheri: I think the popular choice in python for running commands/subprocess is [`invoke`](https://github.com/pyinvoke/invoke)... The author of Nox also [recommended it to me some time ago](https://github.com/theacodes/nox/issues/333)
But I have the impression `invoke`'s use case is much more complex...

src/pyscaffold/shell.py
abravalheri: I am always insecure when it comes to quoting... This code is always invoking `shlex.quote` for POSIX, and I trust `shlex` knows what is best.

Unfortunately `shlex.quote` does not work well for the Windows.

I think I prefer not to quote if I don't have to, because there might be edge cases... (and the topic is complex: https://docs.microsoft.com/en-gb/archive/blogs/twistylittlepassagesallalike/everyone-quotes-command-line-arguments-the-wrong-way)

src/pyscaffold/shell.py
abravalheri: Apparently, using flags in `EDITOR` is a thing:
- https://dev.to/jonasbn/til-integrate-visual-studio-code-with-shell--cli-2l1l
- https://docs.github.com/en/github/getting-started-with-github/associating-text-editors-with-git
- https://stackoverflow.com/questions/8951275/how-can-i-make-sublime-text-the-default-editor-for-git

abravalheri:
It turns out `subprocess.list2cmdline` knows how to escape whitespaces on Windows... so we don't have to implement our own (and we can simplify the code)...

FlorianWilhelm:
Hi @abravalheri, sorry for the really long delay. Judging from what you wrote, especially that `subprocess.list2cmdline` can handle Windows escaping correctly and your code, I think this PR is the way to go, right? Should we merge it and prepare for a new fix release 4.0.2 or wait for 4.1?

abravalheri:
Thank you very much for the review @FlorianWilhelm... I was really not sure which way to go, but I think keeping the shell helps avoid unforeseen side-effects.

I think there will be no harm in having a patch release, specially because the external API is not changing and we are keeping the same behaviour (using shell).

PR #432, opened 2021-03-30
Address deprecation warning from importlib_resources

abravalheri:
In `importlib_metdata` 3.9 and Python 3.10, the `entry_points` function deprecated the `dict`-like interface of its return, instead a new `select` function was introduced.

https://importlib-metadata.readthedocs.io/en/latest/using.html#functional-api


FlorianWilhelm:
Thanks for fixing this warning.

PR #433, opened 2021-03-30
Alternative attempt to solve #430 by passing a list of arguments instead of quoting

abravalheri:
This PR targets #430, similarly to #431, but instead of quoting the file path, this implementation uses a list to represent the command passed to `subprocess.run` and `Popen`. In theory, in this approach, Python should do the escaping itself... 

There might be implications in this change of behaviour... even if the tests pass, it is necessary to check if the extensions are not relying on the ability to pass arguments directly on the `ShellCommand` string (e.g. `ShellCommand("tar xvf")` was previously valid, with this change `ShellCommand("tar", "xvf")` is required).

### DIFF
--- a/tests/test_install.py
+++ b/tests/test_install.py
@@ -169,9 +169,14 @@ class DemoApp(object):
                 #    have to be prepended to PATH, since Windows seems to ship with a
                 #    BSD tar)
             except exceptions.ShellCommandException:
-                untar(self.dist_file)
-                # ^  it seems that some versions of tar for windows are no longer
-                #    supporting --force-local, so let's try without it
+                # It seems that some versions of tar for windows are no longer
+                # supporting --force-local, so we need more debug information
+                print("Path:\n\t", "\n\t".join(os.getenv("PATH", "").split(os.pathsep)))
+                print("Exec path:\n\t", "\n\t".join(os.get_exec_path()))
+                print("tar:", repr(shell.get_executable("tar")))
+                print("gtar:", repr(shell.get_executable("gtar")))
+                print("bsdtar:", repr(shell.get_executable("bsdtar")))
+                raise
 
     def install(self, edit=False):
         with self.guard("installed"), chdir(self.pkg_path):
--- a/tests/test_install.py
+++ b/tests/test_install.py
@@ -32,7 +32,7 @@ __location__ = Path(__file__).parent
 pytestmark = pytest.mark.slow
 
 
-untar = shell.ShellCommand(("gtar" if command_exists("gtar") else "tar") + " xvzkf")
+untar = shell.ShellCommand("gtar" if command_exists("gtar") else "tar", "xvzkf")
 # ^ BSD tar differs in options from GNU tar, so make sure to use the correct one...
 #   https://xkcd.com/1168/
 
--- a/tests/test_install.py
+++ b/tests/test_install.py
@@ -32,7 +32,8 @@ __location__ = Path(__file__).parent
 pytestmark = pytest.mark.slow
 
 
-untar = shell.ShellCommand("gtar" if command_exists("gtar") else "tar", "xvzkf")
+tar = shell.get_executable("gtar" if command_exists("gtar") else "tar")
+untar = shell.ShellCommand(tar or "tar", "xvzkf")
 # ^ BSD tar differs in options from GNU tar, so make sure to use the correct one...
 #   https://xkcd.com/1168/
 
@@ -171,6 +172,7 @@ class DemoApp(object):
             except exceptions.ShellCommandException:
                 # It seems that some versions of tar for windows are no longer
                 # supporting --force-local, so we need more debug information
+                print("untar _command:", repr(untar._command))
                 print("Path:\n\t", "\n\t".join(os.getenv("PATH", "").split(os.pathsep)))
                 print("Exec path:\n\t", "\n\t".join(os.get_exec_path()))
                 print("tar:", repr(shell.get_executable("tar")))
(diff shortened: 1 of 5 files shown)


abravalheri:
From the CI reports for https://github.com/pyscaffold/pyscaffold/pull/433/commits/4dd90abaddd02af06736d11dcc76503aac993425 and https://github.com/pyscaffold/pyscaffold/pull/433/commits/7442bc89a8b3b68271986f6fdc7d9b382389fd15, it seems that `ShellCommand("git")("--version")` is problematic for `subprocess.run` with an argument list and `shell=True` (instead of an argument string, as we used to to previously).

abravalheri:
`shell=False` seems to work better with the approach of passing a list instead of single string to `subprocess.run`. However it does change a bit the semantics, and calling the module/class `shell`/`ShellCommand` without actually invoking the shell by default seems to me counter intuitive 😅 

Ideally a change in approach should also be followed by some renaming...

FlorianWilhelm:
If `shell=False` you also have no access to any environment variable, right?

abravalheri:
I am not sure about that... I suppose you would.

Its been a while I studied this, but environment variables are not strictly associated with a shell, they are a feature of the operating system, like `argc` and `argv`... (my memory might be a bit vague on this topic, but if I am not wrong, every time the OS call an executable it can write in a specific region of the memory a bunch of characters following a convention; using that convention the program can parse this and have access to the values of environment variables/argc/argv).

In Python, I think, as long as you don't pass the `env` keyword argument to `subprocess.run` (or any equivalent `Popen`-based command), the default is for the subprocess to inherit the environment variables of the parent process.

abravalheri:
Here you go (it is not a proper text book, but gives a good idea): https://www.thegeekstuff.com/2012/03/linux-processes-memory-layout/




(so in theory env vars are something that can be done without shell)

abravalheri:
That is a nice way of summarising it 😄 

And yes, the users (in both POSIX/Windows) would always have to manually quote the EDITOR/VISUAL environment variable (if it contains whitespaces) regardless if they are adding or not extra flags.

tests/test_install.py
```diff
@@ -32,7 +32,8 @@
 pytestmark = pytest.mark.slow
 
 
-untar = shell.ShellCommand(("gtar" if command_exists("gtar") else "tar") + " xvzkf")
+tar = shell.get_executable("gtar" if command_exists("gtar") else "tar")
```
abravalheri, line 35: Judging by the results of the CI in the last results: it seems that when `shell=False`, Python has problems to find the correct executable.

It looks like the env variables are available for the children process after it starts, but `PATH` is not being used to find the correct executable when a full path is not provided... (the tests just started working when `shell.get_executable` got introduced here)

abravalheri: I did a little experiment to understand more about this limitation:

On Linux:
```bash
echo -n '#!/bin/sh\necho $MYVAR' > test/tar
chmod +x test/tar
PATH="./test:$PATH" MYVAR=42 python3 -c 'import subprocess; subprocess.run(["tar", "-h"])'
# => 42
```
∴ on Linux, `subprocess.run` is able to find the correct executable considering the current value of `PATH`.

On Windows, I created the files `other.bat` and `tar.bat` inside the `test` folder in my current working dirctory, both containing the following code:
```batch
@echo off
echo %MYVAR%
```
Then I run (on CMD):
```batch
set PATH=test;%PATH%
set MYVAR=42
other
:: => 42
tar
:: => 42
python -c "import os; print(os.getenv('PATH'))"
:: => test;C:\Python38\Scripts\;C:\Python38\; ... ;C:\WINDOWS\system32; ...
python -c "import subprocess; subprocess.run(['other', '-h'])"
:: => ... FileNotFoundError: [WinError 2] The system cannot find the file specified ...
python -c "import subprocess; subprocess.run(['tar', '-h'])"
:: => tar(bsdtar): manipulate archive files ...
python -c "import subprocess; subprocess.run(['tar.bat', '-h'])"
:: => 42
copy C:\WINDOWS\system32\tar.exe .\test\foobar.exe
python -c "import subprocess; subprocess.run(['foobar', '-h'])"
:: => foobar(bsdtar): manipulate archive files ...
rename rename test\foobar.exe notepad.exe
python -c "import subprocess; subprocess.run(['notepad', '-h'])"
:: Opens a notepad window
python -c "import os, subprocess; print(os.getenv('PATH'));  subprocess.run(['notepad', '-h'])"
:: Prints "test;C:\Python38\Scripts\;C:\Python38\; ... ;C:\WINDOWS\system32; ..." but still opens a notepad window
```
∴ on Windows, `subprocess.run` will only find the correct file considering `PATH` if the file extension is given explicitly as part of the file name (e.g. `.bat`), or when it is `.exe` **BUT** executables on `C:\WINDOWS\system32` will have priority, even if the `PATH` is adjusted in a way that `C:\WINDOWS\system32` comes last.

In the case of the failing tests, what was happening is that `C:\WINDOWS\system32\tar.exe` was given priority over `C:\tools\GitForWindows.2.27.0\tools\usr\\bin\tar.exe`

abravalheri:
As indicated in my last comment on the code, there are 2 limitations of using `shell=False` on Windows:

- the executable name will not resolve correctly if the extension is missing (`.exe` seems to be an exception)
- standard Windows locations, such as `C:\WINDOWS\system32`, will have preference when resolving the executable path despite its position in `PATH`.

abravalheri:
@FlorianWilhelm, when you have some time (no need to worry), I would love to hear your feedback about #433, #431 or any other alternative...
This issue is a bit tricky and I am not sure what is the best way forward :P

PR #434, opened 2021-04-01
Move embedding PyScaffold into FAQ

FlorianWilhelm:
## Purpose

Since embedding PyScaffold into another application surely is a topic for a really small percentage of our users, I suggest moving the docs into the FAQ in order to clean up a little bit the sidebar.


abravalheri (approved):
Thank you very much for the improvements @FlorianWilhelm.

PR #435, opened 2021-04-01
Added issue 423 as FAQ

FlorianWilhelm:
## Purpose
This more or less only adds the answers from #423 as FAQ.


abravalheri (approved):
Thank you very much 👍

PR #436, opened 2021-04-01
Larger reorg and simplification of the sidebar

FlorianWilhelm:
## Purpose
From what I hear some users have trouble with PyScaffold's current documentation structure. This is an attempt and suggestion on how to simplify the organization of the docs.


abravalheri:
Thank you very much @FlorianWilhelm, I like the reorganisation, specially the new usage section... It makes a lot of sense to me.

I was wondering, what if instead of putting advanced topics in the FAQ, we actually have an "Advanced" section and then use a toctree to actually include the advanced topics there?

FlorianWilhelm:
Thanks, @abravalheri, that's a good idea! Let me try something...

abravalheri (approved):
I really like it 😄 

Thank you @FlorianWilhelm

FlorianWilhelm:
Thanks @abravalheri :-)

PR #439, opened 2021-04-09
Remove duplicate spec of version_scheme

vladdu:
The version scheme is already specified in pyproject.toml, is it really necessary to duplicate it in setup()?

### DIFF
--- a/src/pyscaffold/templates/setup_py.template
+++ b/src/pyscaffold/templates/setup_py.template
@@ -10,7 +10,7 @@ from setuptools import setup
 
 if __name__ == "__main__":
     try:
-        setup(use_scm_version={"version_scheme": "no-guess-dev"})
+        setup()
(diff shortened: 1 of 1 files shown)


abravalheri:
Hi @vladdu, thank you very much for the PR!

In my mind that will be the default behaviour in v5: rely on the existence of `pyproject.toml`. However, in v4 I left behind a escape hatch with `--no-pyproject`, and then it was just easier to have this definition duplicated.

Are you facing problems because of the duplication?

vladdu:
Hi! Oh, I understand it now. No, the problem is that I always end deleting this from setup.py, so that I don't get confused when changing the versioning scheme in only one place. If nobody else complained, then don't mind me! I suppose it's only going to happen for the first two or three projects one scaffolds :-)

PR #445, opened 2021-05-16
Add one more FAQ about building without git

FlorianWilhelm:
## Purpose

Add one more entry about how to build from source code if no git repo is available.


abravalheri (approved):

abravalheri:

That is very helpful!

Issue #427, opened 2021-03-11
Improve information about conda in docs


abravalheri:
Currently we have [this information available about conda](https://pyscaffold.org/en/latest/conda.html).
According to @maresb review on #424:

> A potential problem I see is that "Creating Conda Packages" is a very niche topic which 99% of users will probably never do. But using conda in conjunction with PyScaffold in particular dsproject is a very important mainstream theme. This niche topic is now very prominent in the PyScaffold documentation, while (as far as I can tell) there isn't really a prominent explanation of using conda in general. I think this gives the wrong impression about what conda is mainly for. I wonder if it would be better to:
> 
> - rename the topic "Creating Conda Packages" → "Using Conda Packages",
> - describe using conda as a package manager, the environment.yml file, etc, and
> - put this section which you just wrote at the end.

Those are really great suggestions that would be nice to adopt.

Hopefully we can borrow some of the previous writings from @FlorianWilhelm on #368 and #426 to compose the extra sections.

Additionally this section should be at least linked to [Dependency Management](https://pyscaffold.org/en/latest/dependencies.html) (should we keep them 2 separated sections or put conda under the Dependency Management? but then talking about how to create conda packages seem a bit weird... not sure how to organise everything in a cohesive way).

FlorianWilhelm:
I'll look into that. I like the idea of putting it into the dependency management section and describe how to use conda instead of creating an own conda package. I'll work on a proposal PR.

abravalheri:
Thank you very much @FlorianWilhelm

FlorianWilhelm:
This was fixed in #446.

PR #446, opened 2021-05-24
Reorg of conda docs

FlorianWilhelm:
This fixes the issue to improve the docs about conda, #427.


FlorianWilhelm:
Oh my gosh, after fiddling more than an hour around with `TOMLDocument` and `Container` datatypes from the new tomlkit version 0.7.2 I am pretty lost. It feels that what they have defined is quite inconsistent and thus impossible to use. Do you have an idea how to solve this @abravalheri? I am kind of lost.

abravalheri:
Hi @FlorianWilhelm, thank you very much for having a look on this.

For the issue with the typecheck, the following is my attempt of organising my thoughts on the topic (sorry if the text is confusing):

### Previous implementation of PyScaffold's `toml.py`

I believe that during the transition from PyScaffold `3 => 4`,  `tomlkit` did not have type hints or were exposing a private type (`_TOMLDocument`, which is the current scenario), so after investigating their code base, my line of thought was:

- `tomlkit.parse` was returning a custom `MutableMapping` subclass, and that was the important part for the rest of the PyScaffold code... we don't need to know exactly which class the return type is, we only need to rely on the `MutableMapping` interface and everything else should be relatively well behaved in terms of type consistency.
- Callables are [covariant](https://blog.daftcode.pl/covariance-contravariance-and-invariance-the-ultimate-python-guide-8fabc0c24278) in terms of return types. So if `CustomClassInTomlKit <: MutableMapping` and we have a function with type `T = Callable[[...], CustomClassInTomlkit]`, `T` is also `Callable[[..], MutableMapping]`, and that is enough for PyScaffold (we can ignore the `CustomClassInTomlKit` as an implementation detail).

I could have stopped there... But I also introduced a semi-[opaque data type](https://en.wikipedia.org/wiki/Opaque_data_type), `TOMLMapping <: MutableMapping`. The idea behind this semi-opaque data type was to *be able to differentiate between python-native MutableMappings such as dicts and MutableMappings obtained via parsing of TOML documents*. Please notice this information is not currently used in the codebase, but in theory could be used to implement more TOML specific functions inside `toml.py` while decoupling from the library interface (which is relatively unstable yet). The main idea is that we replace a type definition that the library controls by a type definition that we control and at the same time workaround of an API described in terms of private data types (?!).
Other advantages are mentioned [here](https://docs.python.org/3/library/typing.html#newtype). If the library changes the types being returned, we adjust the functions inside the `toml.py` and avoid the abstraction leaking to other parts of PyScaffold...


### Current Issues

From what I see in the CI check history, the struggle seem to be happening because `tomlkit` started doing type annotations and now `TOMLDocument` is incompatible with `MutableMapping[str, Any]`.

My opinion is that this is happening because the type hints in `tomlkit` are incomplete/private(?!). According to the specs, [keys should always be interpreted as strings](https://toml.io/en/v1.0.0#keys), but `tomlkit.loads` does not specify the type arguments for `MutableMapping` (in this situation I think mypy assumes `MutableMapping[Any, Any]`). Since `MutableMapping` is invariant and `MutableMapping[str, Any] != MutableMapping[Any, Any]`, the typecheck fails. Moreover `tomlkit.loads` and `tomlkit.dumps` are exposing signatures with a private type (`_TOMLDocument`)

We might have 2 ways around this issue:

1. Change our definition to use `MutableMapping` instead of `MutableMapping[str, Any]`. The theoretical consequence is that we can have more false negatives in the typecheck (uncaught errors). It might also be the case that other parts of PyScaffold rely on the fact that the configuration keys are always strings (we would need to do the change and run the typecheck to be 100% sure).
2. We force our interpretation of the TOML spec with `TOMLMapping(cast(MutableMapping[str, Any], tomlkit.loads(text)))` and add a `# type: ignore[arg-type]` comment on `tomlkit.dumps`. The risk here is that the official spec of TOML changes to accept keys other than strings (less likely).

Other comments:

1. `MutableMapping[str, Any]`, although more precise than `MutableMapping[Any, Any]` is still imprecise in terms of the TOML spec; but implementing a better type definition for the [values](https://toml.io/en/v1.0.0#keyvalue-pair) would be more cumbersome.
2. Sometimes we will be forced to add `cast` or `# type: ignore` anyway, because the Python does not support [dependent types](https://en.wikipedia.org/wiki/Dependent_type). For example, due to the PEP517/518 we know for sure that certain keys in `pyproject.toml` will return ["nested tables"](https://toml.io/en/v1.0.0#table) (interpreted as dicts in Python) or [arrays](https://toml.io/en/v1.0.0#array) (interpreted as lists in Python), while others will return plain strings or numbers. Due to Python's type system limitation, a TOML document will always be interpreted as a tree-like structure, and therefore any item retrieved from a key in `pyproject.toml` will be typed as an union of scalar and compound types, regardless of the definition in PEP517/518.

abravalheri:
What do you think about these options?

If are tired of fighting the typecheck, I could also try to do some extra commits on this branch...

FlorianWilhelm:
Thanks, @abravalheri for the explanation. I think adding an issue describing this strange behavior to tomlkit and using the easiest possible workaround to satisfy our typechecker might be the best solution until tomlkit fixes the problem on their side. That would be option 1, right?

Feel free to push commits directly to this branch. I must say that I am still pretty unfamiliar with everything type-hinting-related in Python. Didn't even know that there is `cast`, good to know.

abravalheri:
So, it seems that either mypy is not recognising TOMLDocument as a subtype of MutableMapping, or NewType creates an invariant constructor... anyway I had to introduce a cast...

I wonder if I should bring the generic type arguments back to MutableMapping, making it `MutableMapping[str, Any]` again (since the cast is already needed).

FlorianWilhelm:
Thanks, @abravalheri! I was also a bit confused about the fact that TOMLDocument is not recognized as a subtype of MutableMapping. 

I have no clue about whether reintroducing the generic type arguments to `MutableMapping` is good or not, to be honest. My suggestion would be to maybe wait until tomlkit has fixed this problem upstream. As far as I see it the interplay of tomlkit's type definitions especially in `dumps`, `loads` and `setdefaults` make no sense at all.

abravalheri:
Thanks @FlorianWilhelm , I agree


I opened a PR yesterday on tomlkit about the private type annotations. If they are opened to PRs I might also add another ones improving the type hints.

abravalheri:
I think we can release... In the unlikely scenario the guys from tomlkit accept the PRs quickly the type annotations here should still work.

PR #448, opened 2021-05-29
Mention Mamba in docs

maresb:
## Purpose

With the closing of #135, I took a final look at the conda docs. It looks very good to me. The one thing which seems to be missing is a mention of `mamba`.

For example, on my computer in a clean `conda` environment, it takes 8 hours to install `dvc` via `conda` but 33 seconds with `mamba`. More details [here](https://github.com/iterative/dvc.org/issues/2370#issuecomment-818891218).

What do you think? Is it good to mention `mamba` and is this a good place?


docs/dependencies.rst
```diff
@@ -314,7 +314,7 @@ If you want to add a dependency to the dev environment only, you can also:
 Integration with conda
 ======================
 
-conda_ is an open-source package manager very popular in the Python
+Conda_ is an open-source package manager very popular in the Python
```
FlorianWilhelm, line 317: Aren't those links case-sensitive?

maresb: Good question. Not on the GitHub preview. But that's probably not the right place to be testing it.

maresb: I built locally, seems to work.

image

FlorianWilhelm: Cool, thanks for checking!

FlorianWilhelm:
Thanks, @maresb, looks good to me. I have never used `mamba` to be honest and didn't know that `dvc` takes 8 hours to install. Maybe I used only the pip version so far. If @abravalheri agrees too, let's merge this :-)

abravalheri:
Thank you very much Ben, that looks very good to me.

@FlorianWilhelm, thanks to @maresb tip, I started using `mamba` a few months ago and it is definetely faster than `conda` (`pip` is still the fatest, but it cannot handle everything the conda ecosystem can...). I think this is a good addition to the docs, and will help people that (like me) are stuck with slow conda installs :P

abravalheri:
This coverage thing is a bit annoying... I remember already seeing it, but it was before we added the `COVERALLS_REPO_TOKEN` env var in the `.cirrus.yml`. By the way, this error is also happening in a recent PR to `configupdater` and in both cases is a docs-only change, which makes things even more weird... It is likely that something happened to coveralls recently.

FlorianWilhelm:
@abravalheri, this coveralls problem seems to appear every time someone other than us, i.e. you and me, make a PR. Hasn't this something to do with the fact that non-maintainers cannot access Github's secret vault or something like that?

Issue #451, opened 2021-06-04
Docs no longer displaying lists correctly


abravalheri:
By the looks of the website, it seems that the latest reST parser used by Sphinx is no longer recognising lists that start at the same level of indentation as their preceding paragraph.
Evidence:
- The third question in the [FAQ](https://pyscaffold.org/en/stable/faq.html)
- the [extensions section in the features page](https://pyscaffold.org/en/stable/features.html#extensions)
- the [Changelog](https://pyscaffold.org/en/stable/changelog.html)

To solve this, looks like adding one level of indentation to the lists is enough (or at least the unordered ones, where I can see the problem).

abravalheri:
Triggering a rebuild on RTD does not solve the problems.

abravalheri:
On further investigation, it seems to be likely a theme bug:

https://github.com/readthedocs/sphinx_rtd_theme/issues/1115

abravalheri:
Closed by #453.

PR #453, opened 2021-06-04
Experiment: Change in the Sphinx Theme

abravalheri:
Recently the RTD themes have been showing some incompatibilities with the latest versions of Sphinx and its dependencies (see #451).

Moreover, instead of centralising the main content in the window, the RTD theme compresses the entire page to the left, and fills the remaining space on the right with a gray background. In large monitors this empty area can have a size comparable (if not bigger) to the area in which the content is displayed, which I consider unpleasant to the reader.

This is an experiment with a new trending Sphinx theme: Furo. It is a minimal theme which prioritises readability and have been used lately in the documentation for the majority of the PyPA projects.

I did a series of customisations/improvements that I found suitable for the PyScaffold docs.

A preview of the changes can be seen [here](https://pyscaffold.org/en/change-docs-theme/).

What do you think @FlorianWilhelm, would that be a good idea?
I do like the original RTD typefaces, but the empty space on the right is a real shame... 

**Note**
I have mainly focusing on the `light` mode of the theme. Depending on the browser/operating system configuration, the `dark` version can be active. If we find this experiment interesting and we want to go forward with the new theme, I can spend some time also changing the `dark` mode.

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -3,12 +3,13 @@ article .align-center:not(table) {
 }
 
 .sidebar-logo {
-  background-color: var(--color-brand-primary);
+  background-color: var(--color-brand-logo-background);
   border-radius: 0.3rem;
 }
 
 main img[alt$="logo"] {
   margin-top: 3rem;
+  background-color: var(--color-brand-logo-background);
   border-radius: 0.5rem;
 }
 
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -1,0 +1,69 @@
+article .align-center:not(table) {
+  display: block;
+}
+
+.sidebar-logo {
+  background-color: var(--color-brand-primary);
+  border-radius: 0.3rem;
+}
+
+main img[alt$="logo"] {
+  margin-top: 3rem;
+  border-radius: 0.5rem;
+}
+
+dl:not([class]) dt {
+  color: var(--color-brand-content);
+}
+
+ol > li::marker {
+  /* font-weight: bold; */
+  color: var(--color-foreground-muted);
+}
+
+blockquote {
+  background-color: var(--color-sidebar-background);
+  border-left: solid 0.2rem var(--color-foreground-border);
+  padding-left: 1rem;
+}
+
+blockquote p:first-child {
+  margin-top: 0.1rem;
+}
+
+blockquote p:last-child {
+  margin-bottom: 0.1rem;
+}
+
+.mobile-header,
+.mobile-header.scrolled {
+  border-bottom: solid 1px var(--color-background-border);
+  box-shadow: none;
+}
+
+.section[id$="package"] h1 {
+  color: var(--color-brand-content);
+}
+
+.section[id^="module"] h2 {
+  color: var(--color-brand-primary);
+  background-color: var(--color-brand-light);
+  border-left: solid 0.2rem var(--color-brand-primary);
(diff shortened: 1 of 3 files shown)


FlorianWilhelm:
Wow, I totally love it! 😍 It feels way more lightweight and modern compared to the old theme. Thanks @abravalheri for the great work.

There seems to be just some small problem with subsections as they seem to appear on the right sidebar, e.g.:
image
Also, is there a way to move the rtd version selector (upper figure at the bottom) to the left side?

Another example when FAQ is selected on the left sidebar, the subsections appear on the right sidebar.
image

Could this be changed so that it resembles the old, more consistent logic?
image

abravalheri:
Hi @FlorianWilhelm thanks for the comments.

The points you mentioned are tricky ones :P. I do agree with them, and I like the right sidebar organisation of the RTD theme better. These are my conclusions after some research:

1. Separated right and left sidebar navigation (instead of a single menu listing both individual pages and sections inside each page):
    According to the theme author, this is a conscious design decision: https://github.com/pradyunsg/furo/discussions/146.
    In order to change that we would need to fork the theme and get deep into the way these sidebars are generated. It is doable, but more of a long term project (at least for me).
2. RTD version menu on the bottom right, instead of bottom left:
    As stated by one of the members of the RTD crew in [a different discussion](https://github.com/python/docs-community/issues/1#issuecomment-836401834), the version switch menu is not part of the theme, but always inserted on the bottom right (except for the official RTD theme). In theory, it is possible to move it via some hacks and abusing the custom.css file.
  The cons of this approach are:
   - Since the RTD version switch menu markup is not documented, the CSS hacks might be unstable
   - The left sidebar menu is collapsable in small screens (e.g. phone), and have some animations (slide right/left) when the menu button is clicked. Since the RTD version switch menu markup is not inserted as a child of the sidebar root element, it is extra hard to make it visually seamless (i.e. make it show/hide together with the collapsable menu). It would also require a lot of hacking and some serious CSS fighting...
   
    There is [some discussion](https://github.com/readthedocs/readthedocs.org/pull/8052) in RTD about providing an API for better theme integration, but nothing is implemented yet.

FlorianWilhelm:
@abravalheri, a kind of expected that, changing a theme is never easy ;-) Forking a theme is not a good option as it will lead to a lot of extra work and thus I would rather live with those two minor downsides of the furo theme.

On the pro side, it really does look a lot better than the RTD theme, so I would say, let's try it out :-) Feel free to merge and thanks a lot for looking into this.

FlorianWilhelm (approved):
LGTM

abravalheri:
Nice! A new theme is exciting!

I suppose it will be easier to customise the RTD version menu once the new API is published.

At some point I was thinking about to fork the theme just to apply the customisations by default for all extensions... but that is much easier. Let's see how it goes (no priority on the theme forking)

abravalheri:

FlorianWilhelm:
BTW, the more I use the furo theme the more I like it, even having the sections and subsections in the right-side-panel and the RTD switcher also.

abravalheri:
😄 @FlorianWilhelm do you see any advantages now? (I still find it a bit weird, but I guess it is a good use of the empty space).

FlorianWilhelm:
On a big screen, the right sidebar becomes more like "compass" that tells you where you are while reading, so it gets updated while scrolling down. This is quite a nice feature a bit like what some IDEs do when scrolling in the source code.

PR #455, opened 2021-06-06
Use READTHEDOCS_VERSION in docs to avoid 'unknown'

### LINKED ISSUE #454: Read The Docs seems to not find correct version
The new theme uses the `version` config to render a top bar on smaller screens.

Unfortunately, it seems that when building on RTD, `version` ends up being equal to `unknown`, as indicated in the image bellow. So I suspect it might be an issue with the `setuptools-scm` integration.

missing version

After looking at the [build logs](https://readthedocs.org/api/v2/build/13938623.txt), I tried to replicated the issue with the following steps:

```bash
$ mkdir -p /tmp/pyscaffold
$ cd /tmp/pyscaffold
$ git clone --no-single-branch --depth 50 https://github.com/pyscaffold/pyscaffold .
$ git checkout --force origin/master
$ virtualenv -p py38 .venv
$ .venv/bin/python -m pip install --upgrade --no-cache-dir pip setuptools
$ .venv/bin/python -m pip install --exists-action=w --no-cache-dir -r docs/requirements.txt
$ cd docs
$ ../.venv/bin/python
```
```python
>>> import sys
>>> sys.path.insert(0, '../src')
>>> from pyscaffold import __version__ as version
>>> version
'unknown'
```

The version is `unknown` even if I run the commands from the root of the project (instead of from inside the docs folder). Apparently the package needs to be installed (e.g. with `pip install -e .`) for `setuptools-scm` to work.

This is the recommendation from `setuptools-scm`:

[setuptools-scm](https://github.com/pypa/setuptools_scm#usage-from-sphinx)

However, without the editable install, it also fails
```python
>>> import sys
>>> sys.path.insert(0, '../src')
>>> from pyscaffold import __version__ as version
>>> version
'unknown'
>>> from importlib.metadata import version
>>> release = version('pyscaffold')
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.8/importlib/metadata.py", line 472, in version
    return distribution(distribution_name).version
  File "/usr/lib/python3.8/importlib/metadata.py", line 445, in distribution
    return Distribution.from_name(distribution_name)
  File "/usr/lib/python3.8/importlib/metadata.py", line 169, in from_name
    raise PackageNotFoundError(name)
importlib.metadata.PackageNotFoundError: pyscaffold
```

Alternatively I tried:
```python
>>> from setuptools_scm import get_version
>>> get_version(root='..', relative_to='conf.py')
/tmp/pyscaffold/.venv/lib/python3.8/site-packages/setuptools_scm/git.py:88: UserWarning: "/tmp/pyscaffold" is shallow and may cause errors
  warnings.warn(f'"{wd.path}" is shallow and may cause errors')
'4.0.3.dev7+g192f609'
```

However this does not considers the `setuptools-scm` configurations in either `pyproject.toml`, or `setup.py`.

abravalheri:
Fixes #454.


docs/conf.py
```diff
@@ -9,14 +9,11 @@
...
 import sys
-import inspect
 import shutil
 
 # -- Path setup --------------------------------------------------------------
 
-__location__ = os.path.join(
-    os.getcwd(), os.path.dirname(inspect.getfile(inspect.currentframe()))
-)
+__location__ = os.path.dirname(__file__)
```
abravalheri, line 16: The previous version is more robust, but also more convoluted.
`__file__` should be enough for the majority of the usages of this particular file.

docs/conf.py
```diff
@@ -46,10 +43,7 @@
 try:
     import sphinx
 
-    cmd_line_template = (
-        "sphinx-apidoc --implicit-namespaces -f -o {outputdir} {moduledir}"
-    )
-    cmd_line = cmd_line_template.format(outputdir=output_dir, moduledir=module_dir)
+    cmd_line = f"sphinx-apidoc --implicit-namespaces -f -o {output_dir} {module_dir}"
```
abravalheri, line 46: Python 3.6 f-strings FTW

abravalheri:
btw, I enabled RTD to build for PRs, so we can also inspect documentation changes.

We can see it now in the checks list as `docs/readthedocs.org:pyscaffold`

src/pyscaffold/templates/sphinx_conf.template
```diff
@@ -281,3 +278,5 @@ intersphinx_mapping = {
     "setuptools": ("https://setuptools.readthedocs.io/en/stable/", None),
     "pyscaffold": ("https://pyscaffold.org/en/stable", None),
 }
+
+print(f"loading configurations for {project} {version} ...", file=sys.stderr)
```
abravalheri, line 282: This line let us inspect in the RTD logs which version is being recognised.

src/pyscaffold/templates/sphinx_conf.template
```diff
@@ -99,11 +93,12 @@ copyright = "${year}, ${author}"
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
 # built documents.
+# The following are default values, that are rewritten later in the script.
 #
 # The short X.Y version.
-version = ""  # Is set by calling `setup.py docs`
+version = os.getenv("READTHEDOCS_VERSION", "")
```
abravalheri, line 99: This env var contains the same kind of string as the ones included in the RTD version switch menu, e.g. `v.4.0.x`, `stable`, `latest`.

docs/conf.py
FlorianWilhelm: Yes, f-strings are awesome.

docs/conf.py
FlorianWilhelm: hm... not so sure about that. You are changing the logic from absolute path to relative path, which might be problematic in some cases. How about:
```
__location__ = os.path.join(os.getcwd(), os.path.dirname(__file__)
```
as a compromise? And another thought. Since version 3.6 the `pathlib` module was already quite evolved. Maybe one can solve this problem already much simpler and nicer with pathlib?

docs/conf.py
```diff
@@ -106,11 +100,12 @@
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
 # built documents.
+# The following are default values, that are rewritten later in the script.
 #
 # The short X.Y version.
-version = ""  # Is set by calling `setup.py docs`
+version = os.getenv("READTHEDOCS_VERSION", "")
```
FlorianWilhelm, line 106: Who is setting this environment variable? I think we should make sure that git tag versions = pypi distributions = the rtd versions.

FlorianWilhelm: I would rather completely remove this part as we set `version` later using `__vesion__`. This only confuses people.

docs/conf.py
abravalheri: It seems that, [starting by Python 3.4](https://docs.python.org/3.4/whatsnew/3.4.html#other-language-changes), `__file__` is absolute.

I also did a quick test here to verify that:
```bash
$ echo "print(__file__)" > /tmp/test_python_file.py
$ python /tmp/test_python_file.py
__file__='/tmp/test_python_file.py'
```

I suppose if you want to use pathlib we could also do:
```python
from pathlib import Path

__location__ = Path(__file__).parent
```

docs/conf.py
abravalheri: So this env variable is set in all environments by Read The Docs.
The reason why I added it here is to have a fallback mechanism when publishing the docs, in the case there is a problems with the package.

docs/conf.py
FlorianWilhelm: good to know that it's absolute now. I am also fine with `os.path.dirname(__file__)` as it might be even easier to read.

docs/conf.py
FlorianWilhelm: Ok, should we add a comment then that it's later potentially overwritten? Or remove it at this point and put it in the try-except block directly? It might be confusing to readers who read `version =` and think, okay so it's an environment variable but later it's overwritten.

src/pyscaffold/templates/sphinx_conf.template
```diff
@@ -162,12 +165,7 @@ html_theme_options = {
 
 # The name for this set of Sphinx documents.  If None, it defaults to
 # "<project> v<release> documentation".
-try:
-    from ${qual_pkg} import __version__ as version
-except ImportError:
-    pass
-else:
-    release = version
+# html_title = None
```
abravalheri, line 168: The preceding original comment was not related to version/release.
Apparently this `html_title = None` was deleted by mistake at some point in the history of the file.

docs/conf.py
abravalheri: I agree with the confusion, in the latest commit to this PR I collapsed both sections of the code manipulating version/release into a single one. Hopefully it avoids confusion.

docs/conf.py
```diff
@@ -107,10 +101,19 @@
...
+# just set them both to the same value.
+try:
+    from pyscaffold import __version__ as version
+except ImportError:
+    version = ""
+
+if not version or version.lower() == "unknown":
+    version = os.getenv("READTHEDOCS_VERSION", "unknown")  # automatically set by RTD
+
+release = version
```
abravalheri, line 116: Now everything is centralised in a single section of the code.

abravalheri:
Thank you very much for the review @FlorianWilhelm. I tried to incorporate your comments in the latest changes, in special:

- Adding an editable install of the package itself to RTD
- Centralise version/release config in `docs/conf.py` in a single region of the code instead of setting it first and later overwriting.

FlorianWilhelm:
Thanks @abravalheri, looks really good!

PR #456, opened 2021-06-06
Fix broken links in the docs

abravalheri:
There are still a few warnings for `tox -e linkcheck`:
```bash
/home/abravalheri/projects/pyscaffold/docs/dependencies.rst.rst:306:
WARNING: broken link: https://anaconda.org
(403 Client Error: Forbidden for url: https://anaconda.org/)

/home/abravalheri/projects/pyscaffold/docs/faq.rst.rst:69:
WARNING: broken link: https://crates.io
(404 Client Error: Not Found for url: https://crates.io/)

/home/abravalheri/projects/pyscaffold/docs/contributing.rst.rst:5:
WARNING: broken link: https://www.blue-yonder.com
(HTTPSConnectionPool(host='www.blue-yonder.com', port=443): Max retries
exceeded with url: / (Caused by
SSLError(SSLCertVerificationError("hostname 'www.blue-yonder.com'
doesn't match either of '*.jda.com', 'jda.com'"))))
```

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -205,7 +205,7 @@ coverage_task:
 
 linkcheck_task:
   name: linkcheck (Linux)
-  only_if: $CIRRUS_BRANCH == 'master'
+  # only_if: $CIRRUS_BRANCH == 'master'
   allow_failures: true
   clone_script: *clone
   container: {image: "python:3.8-buster"}
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -201,3 +201,19 @@ coverage_task:
   <<: *REGULAR_TASK_TEMPLATE
   coverage_script:
     - coveralls
+
+
+linkcheck_task:
+  name: linkcheck (Linux)
+  only_if: $CIRRUS_TAG =~ '^v.*'
+  allow_failures: true
+  clone_script: *clone
+  container: {image: "python:3.8-buster"}
+  depends_on:
+    - test (Linux - 3.8)
+  pip_cache: *pip-cache
+  tox_install_script:
+    - python -m pip install --upgrade pip setuptools tox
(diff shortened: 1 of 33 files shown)


FlorianWilhelm:
@abravalheri thanks a lot for fixing all those broken links!

Regarding the remaining ones:

* I think it's now https://blueyonder.com/ instead of with a dash or was there ever www.blue-yonder.com, @BY-jk?
* https://anaconda.org/ works for me
* https://crates.io/ also works for me. That's the repo of rust programmers and should be quite reliable.

abravalheri:
Thanks @FlorianWilhelm for the review.

By removing the hyphen in `blue-yonder` that warning is silenced :)

I forgot to mention in the text of the PR: 

- I tested all the links manually and they seem to work... it might be just an odd behaviour because the URLs are being accessed with a library instead of the browser... Who knows.
-  I tried to convert a lot of `latest` RTD links to `stable` (it makes sense right?), but I there is plenty more. At least they are not broken.

FlorianWilhelm:
Thanks, @abravalheri, yes, it makes sense to convert `latest` to `stable` as this should always refer to the latest stable version instead of the latest development.

The PR looks good to me.

abravalheri:
I corrected a bunch of other URLs and I added a CI task (with `allow_failures: true`) to run on the tags, so at least we should be aware when links are broken...

Unfortunately we will have to test manually this extra task by triggering a tag...

BY-jk:
> @abravalheri thanks a lot for fixing all those broken links!
> 
> Regarding the remaining ones:
> 
> * I think it's now https://blueyonder.com/ instead of with a dash or was there ever [www.blue-yonder.com](http://www.blue-yonder.com), @BY-jk?

That is correct.

PR #460, opened 2021-06-19
Workaround for links in the left sidebar on docs

abravalheri:
The issue and the implemented workaround are discussed in https://github.com/pradyunsg/furo/discussions/155


FlorianWilhelm:
Hi @abravalheri, just took a look at the RTD of this branch, now it looks like:

image

Now it's really deeply nested and the other points are missing making the `Examples` somewhat detached in the context. As you already discussed with the maintainer of furo that somehow it's not possible to just not add a dropdown for `Extending PyScaffold`, maybe one could use your `:doc:` workaround don't do:
```
.. toctree::
   :hidden:

   examples/index.rst
```
Wouldn't in this case the dropdown menu disappear in the left sidebar but we would still have `Examples` on the right?

abravalheri:
Thank you for the review @FlorianWilhelm, I agree with you about the nesting.

The main issue here is making Sphinx understand it needs to build those files too... I suppose we could use the [`:orphan`: directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html), but that smells like a hack to me :P .

I will try it out.

abravalheri:
Yeap, using the `:orphan:` hack I was able to remove the Examples section from the left sidebar menu:
https://pyscaffold--460.org.readthedocs.build/en/460/extensions.html

I call it hack because some tutorials, such as [the anaconda one](https://docs.anaconda.com/restructuredtext/index.html) advise to avoid using it.

abravalheri:
Sorry @FlorianWilhelm, I think I have previously misunderstood your advice 😅 

I thought you have mentioned the  `:hidden:` in the `toctree` inside `docs/extensions.rst`. I tried that and it was not working... But this morning after re-reading your comment I realised you were mentioning a `:hidden:` `toctree` inside `docs/index.rst`, and that actually works without the need for any `:orphan:` statement!

Thank you very much 🎉

abravalheri:
Yeap, there is no other way around...

If I add a `:hidden:` `toctree` to` docs/index.rst`, the titles of the examples still show in the navigation menu, they are just hidden from the in-page rendering of the `toctree` as a bullet list.

I can see that according to https://github.com/pradyunsg/furo/discussions/117, the philosophy is to show everything everywhere, so in my latest commit I reverted back to using `:orphan:`, which apparently is the only workaround.

FlorianWilhelm:
Thanks @abravalheri, looks good.

PR #462, opened 2021-06-19
Add 0BSD license to options

abravalheri:
Now that PyScaffold itself includes the 0BSD license, it would be fair to also present it as an option.

I included a little title (also for the BSD-2 and BSD-3), as it appears in choosealicense.com.


FlorianWilhelm:
BTW, @hvhaugwitz pointed me to the website https://tldrlegal.com/ which is much more organized than https://choosealicense.com/

abravalheri:
That is a nice one for the info. Well choosealicense is still easier to copy :P

FlorianWilhelm:
Thanks @abravalheri!

PR #463, opened 2021-06-19
Update copyright notice

abravalheri:
Hi @FlorianWilhelm, what do you think about this change? This is just a suggestion that we can simply drop (I am not a lawyer, just replicating patterns 😝 I see around)... 

Since we are updating PyScaffold's license, I thought about updating the copyright notice also to reflect the current situation of the project.

The main inspiration comes from projects like:

- https://github.com/ponylang/ponyc/blob/main/LICENSE
- https://github.com/Homebrew/brew/blob/master/LICENSE.txt

Alternatively we could do like NetBeans does and describe the transferring more in details:

- https://github.com/apache/netbeans/blob/master/NOTICE

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -7,7 +7,8 @@ projects, are licensed under BSD 0-Clause license; see below for details.
 
 The MIT License (MIT)
 
-Copyright (c) 2014 Blue Yonder GmbH
+Copyright (c) 2018-present, PyScaffold contributors
+Copyright (c) 2014-2018 Blue Yonder GmbH
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
@@ -31,7 +32,8 @@ SOFTWARE.
 
 BSD 0-Clause License
 
-Copyright (c) 2014 Blue Yonder GmbH
+Copyright (c) 2018-present, PyScaffold contributors
+Copyright (c) 2014-2018 Blue Yonder GmbH
 
 Permission to use, copy, modify, and/or distribute this software for
 any purpose with or without fee is hereby granted.


abravalheri:
Other references:

- https://github.com/prometheus/prometheus/blob/main/NOTICE
- https://github.com/twbs/bootstrap/blob/main/LICENSE

FlorianWilhelm:
Thanks @abravalheri, I second this! What does @BY-jk as Blue Yonder OSS Evangelist think about this?

BY-jk:
Changing a license can come in two flavors:

* changing the license for new contributions
* changing the license for the overall project

In any case, it is a tricky operation that should be well-thought through. This blog post gives some guidance. https://opensource.guide/legal/ (especially chapter 6).
 
I check the history and noticed that you already changed the license. You might have jumped the gun here. 

If I read the git history correctly, the license for files under pyscaffold.templates was changed from MIT to BSD-0. This is a change to a more permissive license. This can be done, if all copyright holders agree. The easiest way to be compliant is to ask all copyright holders (git history) for consent and document that consent.

I assume the intent is to ensure that files generated from templates are not in any way tainted by Pyscaffold's license. If so, I would assume that all copyright holders would consider this appropriate and thus will consent.

FlorianWilhelm (approved):

FlorianWilhelm:
Thanks @BY-jk, so if it's fine by you, then let's just merge this.

BY-jk:
As I wrote, on the change of license you were jumping the gun on this change: https://github.com/pyscaffold/pyscaffold/pull/462). 

I do not see myself in the position to approve or disapprove. But my recommendation is to circle back with the copyright holders (e.g. all the individuals) that contributed to the templates in pyscaffold.templates. That way, you could ensure that the change to BSD-0 for that part of the code base is approved by all original contributors that are holding the copyright.

PR #465, opened 2021-07-01
replace py.test with pytest

xg1990:
## Purpose

`py.test` is outdated and should be replaced with pytest

> pytest used to be part of the py package, which provided several developer utilities, all starting with py., thus providing nice TAB-completion. If you install pip install pycmd you get these tools from a separate package. Once pytest became a separate package, the py.test name was retained due to avoid a naming conflict with another tool. This conflict was eventually resolved, and the pytest command was therefore introduced. In future versions of pytest, we may deprecate and later remove the py.test command to avoid perpetuating the confusion.

## Resources & Links

Related posts:

    https://stackoverflow.com/questions/39495429/py-test-vs-pytest-command
    https://meta.stackoverflow.com/questions/366905/rename-py-test-to-pytest?noredirect=1&lq=1

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -116,7 +116,7 @@ pyscaffold.cli =
 # e.g. --cov-report html (or xml) for html/xml output or --junit-xml junit.xml
 # in order to write a coverage file that can be read by Jenkins.
 # CAUTION: --cov flags may prohibit setting breakpoints while debugging.
-#          Comment those flags to avoid this py.test issue.
+#          Comment those flags to avoid this pytest issue.
 addopts =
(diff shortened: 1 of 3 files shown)


abravalheri:
Thank you very much @xg1990. That is a very welcome update.

I suppose [the tox template](https://github.com/pyscaffold/pyscaffold/blob/master/src/pyscaffold/templates/tox_ini.template) already uses the more modern executable name, right? So the packages being generated are already correct, but the docs needed the update, is that the case?

xg1990:
Hi @abravalheri , 
Yes I have check the whole source code and it's just a doc update
Cheers, :)

Issue #376, opened 2021-01-04
Add a CONTRIBUTING.rst to projects set up with PyScaffold


FlorianWilhelm:
## Describe your use-case
In general, it seems like a common practice to have a `CONTRIBUTING.rst` file in the root of each repo. This helps users to get started. Also Github's Settings -> Community Option checks for the presence of one. 

## Describe the solution you would like to see for your use-case
One could add a default `CONTRIBUTING.RST` and also reference this file from the Sphinx index. Github provides several examples and there are [others](https://github.com/audreyfeldroy/cookiecutter-pypackage/blob/master/%7B%7Bcookiecutter.project_slug%7D%7D/CONTRIBUTING.rst) on the net.

abravalheri:
I have added a series of PRs to address this issue, by revamping PyScaffold's own `CONTRIBUTING.rst` (#468 and #471) and then using it as a reference to create a template for new projects (#470).

There were a few things that I was not 100% sure:
- In order to improve `CONTRIBUTING.rst`, I have created a dev-guide (#467), but it shares a lot of content with the extension documentation. The most correct solution in this situation would be changing the extension documentation to remove the repeated content and just refer using a link to the dev-guide. However, this solution might be confusing for extension creators (they don't need to know how PyScaffold internal architecture works). Moreover, if the core concepts are not in the same page as the extensions, readers might be inclined to skip them... The solution I found was to separate the repeated code (#466) and use a `include` statement in sphinx to avoid double bookkeeping and inconsistency. *This might not be ideal*
- The proposed template depends on a series of choices from the package developer (e.g., where the code is being hosted, if they are using a public package index, etc). The way I worked around this issue was to add a series of `TODO` items to the template guiding the user to review it before compiling the docs (I also added a configuration to make Sphinx raise a warn for every TODO statement left in the docs).
- I tried to make the proposed template generic for open source and private projects, but I don't really know if it reads nicely for private projects...

Any suggestions/reviews are welcome.

abravalheri:
@FlorianWilhelm I think we are all set here right?

If that is the case, I was planning to release v4.1...

So far, we have accumulated as new features/changes: a new `linkcheck` task in tox.ini, updated licenses (0BSD for generated code + new tamplate) and the new `CONTRIBUTING`.

Should we wait a bit for something else? I think the #447 can wait another minor release...

FlorianWilhelm:
[...]

PR #466, opened 2021-07-04
Split extensions.rst into re-usable files and add pipeline image (1st step #376)

abravalheri:
This is the first pull request designed to address #376.

The main idea is to start by revamping PyScaffold's own contributors guides to use it as a reference for the template.

It is recommended for any contributor's guide to include a reference to the overall description of the project architecture and some guidance about how the project works. Therefore this PR refactors the existing `docs/extensions.rst` into re-usable portions that can be included in the dev guides.
I also added a picture for the internal action pipeline.

### DIFF
--- a/docs/action-pipeline.rst
+++ b/docs/action-pipeline.rst
@@ -1,0 +1,59 @@
+Action Pipeline
+===============
+
+PyScaffold organizes the generation of a project into a series of steps with
+well defined purposes. As shown in the figure bellow,
+each step is called **action** and is implemented as a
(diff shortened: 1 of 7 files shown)


abravalheri:
Since GitHub does not support dependencies between PRs, what I had to do is to set them in a way the following PR targets the branch of the preceding one instead of the master (so the diffs are displayed properly).

This means that for merging, the best would be accepting them in the reverse order.

FlorianWilhelm:
Thanks @abravalheri, looks good to me!

PR #467, opened 2021-07-04
Add a dev guide to docs, to help potential contributors  (2nd step #376)

abravalheri:
This PR depends on #466 and is the second step towards #376.

Any improvement suggestions are welcome. I decided to replicate the entire sections about the core concepts in both `docs/dev-guide` and `docs/extensions`, since extension developers don't need to know how PyScaffold works internally, and it might be confusing for them if we simply include a link pointing out to a more complex process. But I admit it is not ideal...

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -23,7 +23,7 @@ reference in terms of community guidelines.
 
 .. tip::
    If you are new to open source or have never contributed before to a software
-   community, please have a look at `contribution-guide.org`_ and the
+   project, please have a look at `contribution-guide.org`_ and the
    `How to Contribute to Open Source`_ guide.
    Other resources are also listed in the excellent `guide created by FreeCodeCamp`_.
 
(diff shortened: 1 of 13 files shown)


FlorianWilhelm:
Oh sorry @abravalheri, with reversed order you meant steps 1, 2, 3 or 3, 2, 1? I started with 1 and now there are conflicts?

abravalheri:
Sorry @FlorianWilhelm , this system is very complicated :P
It is just that otherwise Github will display the "accumulated changes" in the PRs which make the review more difficult (we have to come up with/agree on a better system for that 😝)

You can review (if you have time), the changes in order (1, 2, 3, 4), but the merging would be facilitated if we do reversed (4, 3, 2, 1). Something that you can also do is approve/reprove the changes an then I take care of merging them myself in a way that no conflicts are found.

FlorianWilhelm:
Alright, stupid me, so step 1 is based on master but step 2 on step 1, step 3 on step 2 etc. so that it needs the reverse order, got it. I'll just click approve then, sorry for the extra work. Can we easily restore the mess I created?

abravalheri:
No problems, it should not be very problematic.

Thank you very much for the review @FlorianWilhelm .

(There will be more PRs related coming :P, one adding the template and another one with complementary changes to Pyscaffold's own contributing.md with some research I did tonight)

FlorianWilhelm:
lgtm

PR #469, opened 2021-07-04
Add PyScaffold badge

FlorianWilhelm:
Add a little PyScaffold badge to the README.


abravalheri:
Amazing!

Thanks for looking into that Florian, it is a very nice visual aid.

Should also we add a badge to the generated README (currently we have a footer saying "Project generated with PyScaffold")?

FlorianWilhelm:
Cool idea. Then we should also do it for the markdown and dsproject extension.

PR #468, opened 2021-07-04
Revamp contributing guides (3rd step #376)

abravalheri:
This PR depends on #467 and is the third step towards #376.

Here I focus in improving PyScaffold's own contributing guides that now include a mention to the dev-guide (which explains how PyScaffold works internally).

It also tries to make it more beginner-friendly by mentioning activities other than coding that can also help the project.

PR #470, opened 2021-07-06
Add contributing templates to generated package (4th step #376)

abravalheri:
This PR depends on #468 and is the 4th step towards #376.

In particular the new `CONTRIBUTING.rst` template is added to newly generated projects.

I tried to keep the tone of the file neutral in a way that it could be used by both open source and proprietary projects, but that is a bit challenging. Moreover, some details of the `CONTIBUTING.rst` depend on the development stack (e.g., if the code is being hosted in GitHub, if the package index is PyPI and etc), so I added quite a few `TODO` notes throughout the document. I am not exactly sure about this approach, so any suggestion is welcome.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -23,7 +23,7 @@ reference in terms of community guidelines.
 
 .. tip::
    If you are new to open source or have never contributed before to a software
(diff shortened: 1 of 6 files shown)


FlorianWilhelm:
Thanks, lgtm.

Issue #473, opened 2021-07-06
Should we add a PyScaffold Badge to the generated README's?


abravalheri:
This is a follow up on #469.

We currently have a footer saying "Project generated with PyScaffold", maybe we should also add a badge on top of the page? People love visual stimulation, and badges are cool.

FlorianWilhelm:
Thanks, @abravalheri, let's close this issue when we have it also in the markdown and dsproject extension. I'll take a shot at it if I find some time over the weekend.

FlorianWilhelm:
As this is also done now in `pyscaffoldext-dsproject` and `pyscaffoldext-markdown`, we can close this issue.

abravalheri:
Thanks @FlorianWilhelm

PR #474, opened 2021-07-06
Add PyScaffold badge to generated projects

abravalheri:
This targets #473 and also add a few examples of other badges that might be used.


FlorianWilhelm:
Thanks, @abravalheri!

Issue #449, opened 2021-05-29
Problems in CI when uploading data to coveralls


abravalheri:
Recently we started noticing errors in the CI tasks during the upload to coveralls.
This can be seen in #448 and https://github.com/pyscaffold/configupdater/pull/24.

The weird part is that both repositories have `COVERALLS_REPO_TOKEN` configured in `.cirrus.yml`.

abravalheri:
We can see that a new version of the Python package for coveralls was published a few days ago:
https://pypi.org/project/coveralls/#history

---
# Things that might help

Coveralls mention a list of extra env variables that we can try to set in the cirrus-ci:
https://docs.coveralls.io/supported-ci-services

```
CI_NAME
CI_BUILD_NUMBER
CI_BUILD_URL
CI_BRANCH
CI_PULL_REQUEST (optional)
```

Coveralls Python CLI mentions:

```
coveralls debug
```

abravalheri:
Thanks @FlorianWilhelm for pointing out what seems to be the problem in #448. Regular contributors cannot access Cirrus CI secrets. 

This behaviour seems to make sense, since PRs could expose secrets, but it is not very handy is it? :P

I suppose the alternative would be exposing a plain text, but [coveralls docs](https://docs.coveralls.io/supported-ci-services) seem to recommend keeping the token secret:

> The option repo_token (found on your repository’s page on Coveralls) is used to specify which project on Coveralls your project maps to. This is only needed for private repos and should be kept secret – anyone could use it to submit coverage data on your repo’s behalf.

Not very sure if we should just ignore these errors when they happen. But it would be nice to know how much a PR changes in terms of coverage.

abravalheri:
I opened [issues in coveralls-python](https://github.com/TheKevJames/coveralls-python/issues/299) and in the [support repository of coveralls itself](https://github.com/lemurheavy/coveralls-public/issues/1557) asking for some guidance about this problem.

abravalheri:
According to the [discussion with the Cirrus-CI maintainer](https://github.com/cirruslabs/cirrus-ci-docs/discussions/866), using shared secrets in PRs is a bit tricky.

Currently, we use the `Only users with write access` policy for decrypting secure env vars. We could change this setting to `Everyone` but then, there is a risk of non-well intentioned PRs change the CI configuration files to expose the secrets in plain text.

[...]

PR #479, opened 2021-07-29
Remove explicit COVERALLS_REPO_TOKEN

abravalheri:
… rely, instead, on environment variable overwritten by Cirrus GUI

This should fix issue #449.
As discussed in the issue, the idea is to use the GUI to provide the env var, which [should be readable by all contributors](https://github.com/cirruslabs/cirrus-ci-docs/discussions/866#discussioncomment-1051665)

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -193,7 +193,6 @@ coverage_task:
   container: {image: "python:3.6-buster"}
   env:
     COVERAGE: yes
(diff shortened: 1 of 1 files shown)


abravalheri:
The approach discussed in #449 seems to work...  🤞 
We are only going to know when a new PR from other contributors arrive...

FlorianWilhelm:
Cool, thanks. LGTM :-)

PR #478, opened 2021-07-28
Add gif to usage docs

abravalheri:
Related to the discussion in #477 .

I think once the PR is merged into master and deployed in RTD, I can also try to add it to the README.

[This is how it is currently looking like](https://pyscaffold--478.org.readthedocs.build/en/478/usage.html)

I have also added a link for a slightly longer version on asciinema.

PR #480, opened 2021-07-30
Add demo gif to README

abravalheri:
This is a follow up on #478. Now that the gif is published in RTD, we can use the image URL on the README.

I also notice that I did a mistake 😝. The document mentions `my_project`, but I have recorded the demo with `myproj`... If editing the gif was not so painful I would change it, but that is a task for the future...

Meanwhile I tried to de-emphasize the name difference to avoid adding confusion to the text. In practice I removed the direct references to `myproj` in the paragraph preceding the gif.


FlorianWilhelm:
Thanks @abravalheri!

PR #491, opened 2021-09-06
Issue 483

### LINKED ISSUE #483: Inconsistent versionning from git tags

I use `pyscaffold` for several Python projects (amazing work btw!). On one of them with `pyscaffold 4.0.2`, I am building with `tox -e build` and generate `sphinx` documentation with `tox -e docs` with theme from `readthedocs`. The version infered from commit tag is OK for the build, but never for the documentation build: the displayed version on the html page is always `0.0.0`.  
Besides, when I type `python setup.py --version` the result is correct, but when doing the following in the Python console:

```bash
from mypackage import __version__ as version
print(version)
```

The result is also falling back to `0.0.0`.

## Describe the solution you would like to see for your use-case

Replace in generated `setup.py`:

```python
setup(
            use_scm_version={"version_scheme": "no-guess-dev"}
        )
```

by:

```python
setup(
            use_scm_version={"version_scheme": "no-guess-dev"},
            setup_requires=["setuptools_scm"],
        )
```

In my case, it solves the issue. It seems to be the way to go in [`setuptools_scm`](https://github.com/pypa/setuptools_scm) documentation.

* PyScaffold Version: 4.0.2
* Python Version: 3.8.10
* Operating system: Ubuntu 20.04.2.0 LTS
* How did you install PyScaffold: conda

abravalheri:
This PR should fix #483. As described in the issue the solution proposed is to use `isolated_build` also for `docs` tasks with tox.

It seems that recent versions of tox just accept the `isolated_build` option as a global config and not per-environment, which is consistent with their most updated docs.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,7 +22,7 @@ Version 4.1, 2021-09-XX
 - Added 0BSD license template
 - Added ``CONTRIBUTING.rst`` template, :issue:`376`
 - Added PyScaffold badge to ``README`` template, :issue:`473`
-- Updated Cirrus CI config and templates, including better ``coveralls`` integration, :pr:`449`
+- Updated Cirrus CI config and templates, including better ``coveralls`` integration, :issue:`449`
 - Adopted global ``isolated_build`` for ``tox`` configuration, :issue:`483`, :pr:`491`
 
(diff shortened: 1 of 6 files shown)


FlorianWilhelm:
Cool! Thanks, @abravalheri

Issue #447, opened 2021-05-26
Support PEP 561 - Typing Support


CarliJoy:
## Describe your use-case
Newer MyPy Version require packages to support [PEP 561](https://www.python.org/dev/peps/pep-0561/) to be considered typed.
It would be great if there is switch to enable typing support for pyscaffold
Next to supporting PEP 561 it would be great if i.e. tox / CI/CD scripts will include also a stage for type checking with mypy

## Describe the solution you would like to see for your use-case
Adding a switch `--typed` that will at least create the `py.typed` file in the `src` folder.
Next to this also the .tox configuration is adopted.

Do you see such a switch within pyscaffold or more within an extension?
Including it within the default pyscaffold will allow other extension writers i.e. of CI/CD scripts to also adopt their pipelines accordingly.

abravalheri:
Hi @CarliJoy, thank you very much for the info... I was really not away that mypy started requiring `py.typed` to be present.

I think for this particular issue, making it an internal implementation might make more sense (this way other extensions could check a `typed: bool` flag in options and also generate typed files...). But we can discuss that, as always there are pros and cons.

For the sake of brainstorm, these are a few snippets that can be useful when implementing this feature (taken from a [PR](https://github.com/pyscaffold/configupdater/pull/23/files) I made to `configupdater`):

```ini
# tox.ini

[testenv:typecheck]
description = invoke mypy to typecheck the source code
changedir = {toxinidir}
passenv =
    TERM
    # ^ ensure colors
deps =
    mypy
commands =
    python -m mypy src
```

```ini
# mypy configuration (setup.cfg)

[mypy]
pretty = True
show_error_codes = True
show_error_context = True
show_traceback = True
ignore_missing_imports = True
warn_redundant_casts = True
warn_unused_ignores = True
```

```yaml
# .cirrus.yml

typecheck_task:
  name: typecheck (Linux - 3.9)
  container: {image: "python:3.9-buster"}  # most recent => better type support
  pip_cache: *pip-cache
  mypy_cache:
    folder: .mypy_cache
  install_script: &debian-install
    - apt-get install -y git
  mypy_install_script:
    - python -m pip install --upgrade pip setuptools mypy -e .
  typecheck_script:
    - python -m mypy src
```

My question is: from the user perspective, when someone uses `--typed`, would they expect all the existing extensions to generate typed code? (e.g. when using `pyscaffoldext-django`, `pyscaffoldext-dsproject`, etc?)

[...]

Issue #336, opened 2020-12-14
Move isort settings to pyproject.toml [4.x]


CarliJoy:
## Describe your use-case
There is a deciated isort.cfg file but no other tool (besides pre-commit) has its one config file, thats confusing.

Also isort.cfg only get added if --pre-commit is used. 

## Describe the solution you would like to see for your use-case
Add the isort.cfg into `pyproject.toml` or `setup.cfg` (both possible according to https://pycqa.github.io/isort/docs/configuration/config_files/)

Then it would be also okay (IMO) to include the settings by default in the pyproject.toml because they are related to the other tools (black for instance) used by the template.

IMO that would be the best and easiest solution:
- remove the isort.cfg and move the content to the default pyproject.toml

CarliJoy:
I could create a PR if I you are okay with the procedure @FlorianWilhelm

abravalheri:
Hi @CarliJoy, thank you for bringing this discussion forward. I agree with you that the way things are right now is definitely not the most "uniform" or "ideal" way of doing things, however I would argue that it is the more pragmatic 😝

For the version 4, I think we are still committing to provide people a escape hatch from isolated builds so they can still do some old fashion packages while things are not stable enough (via `--no-pyproject`). In that case, working conditionally with the flags and trying to assertively position the configuration for isort would result (in my opinion) in added complexity.

We could also try other alternatives like using `setup.cfg` or `tox.ini`, but then the future of these files are uncertain...

By the way, we do have some individual files for other tools like `tox.ini`, `.coveragerc`, `.travis.yml`, `.cirrus.yml`. Do you think `.isort.cfg` is particularly confusing?

PR #493, opened 2021-09-09
Add Typing Support (PEP 561 - issue #447)

abravalheri:
The changes here try to address #447, but they would be more of a RFC than a PR...

I tried to add the most complete implementation that would support the use case described, however there might be room for simplification (currently it is very complex).

One of the things that I noticed is that this implementation highlights problems related to #336, and exposes an abstraction leak if we move in that direction: *the project structure abstraction works very well for updating files, but the mental model is not easily applicable to updating **parts of files** or composing them incrementally with different extensions*.

In the code provided we can see this abstraction leak by comparing the `add_typing_support` and `update_existing_config` actions: they are very similar. The fact that we need 2 different actions that are alternating most of times is also a symptom. The `modify_file` is basically doing the same as `structure.modify_contents` but instead of operating in the dict representation, it operates directly on the file system.

---
Simplifications we could do, and reasons why I haven't implemented that way (yet):

1. We could add a `typecheck` task to `tox.ini` by default, since running tox is *opt-in*.
    - In #447, it is mentioned that other extensions could add type dependencies (e.g. `pyscaffoldext-django` and https://github.com/typeddjango/django-stubs). The way I implemented it was by exposing a `typecheck_deps` option that can be augmented when extensions are activated. The logic to add these deps to `tox.ini` by default would be very similar to the one implemented and therefore I don't see many simplifications by making it a default behaviour.
    - The current implementation have the advantage of being able to update existing project.

2. We could add the mypy config to `setup.cfg` by default.
    - I have the impression that doing so have the potential to mess up the behaviour of some IDEs in the case the user does not want to use typing.
    - Some type dependencies (e.g. the `django-stubs` previously mentioned) require mypy to have some plugins configured, and therefore we would need to have some level of manipulation of `setup.cfg` anyway.
    - The current implementation have the advantage of being able to update existing project.

3. We could add type hints to `skeleton.py` by default.
    I have chosen to not do so because some Python beginners coming from other languages might feel overwhelmed by a type system.

---
Things that I dislike in the PR implementation:
- `retype` being an optional dependency that is required to run `--type`. The alternative would be either making `Typed` an external extension or making `retype` a default dependency...

Things that I like in the current implementation:
- `structure.modify_contents` seems to be a nice addition to the API.


FlorianWilhelm:
Thanks for the suggestions! Sorry for the late reply and I also only had a quick look at the changes. Right now got I lot of work on my plate as colleagues are on vacation.

I must say that it really looks a bit complicated and adds more complexity in the PyScaffold codebase for a feature I am not 100% convinced of. Option 3, you are mentioning, actually sounds like the best for me. This is for several reasons. Most Python programmers I work with are adopting types in their code directly. And with PyScaffold we are addressing clean code affine and thus more advanced developers anyway, not beginners that just start scripting. So for our target group seeing some types shouldn't be a problem and I strongly believe typed Python code will become the default in production-grade code. Since PyScaffold helps you to develop production-grade packages, why not have a template example that is typed by default? Option 3 would be simple, with no need for an extension and extra complexity in the code-base.

What do you think?

abravalheri:
Thank you very much for the review @FlorianWilhelm, I implemented a few changes suggested:

- Types are added by default to `skeleton.py`
- `typecheck` task added by default to `tox.ini`
- `mypy` config added by default to `setup.cfg`
- `typecheck` task added by default to `.cirrus.yml`, however it is disabled via an ENV var working as a flag (otherwise people that do not opt-in for `--typed` will have their build failing).

The remaining complexity is due to:
- **a.** The ability to "flip" the `TYPE_CHECKING` flag on `.cirrus.yml` on `--type` (and therefore enable the task)
- **b.** The ability to add typechecker config to `tox.ini/setup.cfg` in existing projects via `--update`
- **c.** The ability of having other extensions adding dependencies for the typecheck (`django-stubs` is mentioned in #447)
- **d.** The related ability of adding `mypy` plugins (that come with the typecheck dependencies) to the configuration.

I have not added the ability to add a typecheck task to `.cirrus.yml` when updating existing projects since that would require us parsing YAML and therefore another dependency.

It is up for discussion if we don't want to support **b**, **c** or **d**, but I have the feeling that we don't have much option for **a** (unless we opt for  *not having* a `--typed` flag).

abravalheri:
Thank you very much for the suggestions @CarliJoy!
Yeap, we can go for a "non-breaking"/just-informative CI action. What do you think @FlorianWilhelm.

In the future, what I was thinking is:

- Add a `pyscaffold.file_modifiers` module to help with the issue of adding/replacing/removing/commenting-out parts of the text in existing files.
  This seems to be a "missing utility" for extensions (e.g. pyscaffoldext-markdown adds stuff to `docs/config.py`, the `typed` extension could comment/uncomment parts of the `.cirrus.yml` file).
- Rely on [`.cirrus.star`](https://cirrus-ci.org/guide/programming-tasks/) and create a shared starlark library to do autodiscovery of some tasks (e.g. just run the type check if a person have a `py.typed` file).

Of course, those things need to be debated and mainly I need some time to develop them, but I think it could be cool.

FlorianWilhelm:
Thanks @CarliJoy, I like your solution. So typing by default, a switch for the checks in PyScaffold and in total not much more complexity. Sounds really good :-)

abravalheri:
@FlorianWilhelm, do you prefer always performing the typecheck in Cirrus and just failing silently, or use a environment var and `only_if: $TYPE_CHECKING == 'true'` like it is implemented in this PR?

(Both are basically changing a switch)

FlorianWilhelm:
> In the future, what I was thinking is:
> 
> * Add a `pyscaffold.file_modifiers` module to help with the issue of adding/replacing/removing/commenting-out parts of the text in existing files.
>   * This seems to be a "missing utility" for extensions (e.g. pyscaffoldext-markdown adds stuff to `docs/config.py`, `pre-commit` currently modifies the `README`, and the `typed` extension could comment/uncomment parts of the `.cirrus.yml` file).

Will `pyscaffold.file_modifiers` use the functionality of other libraries like `pyaml`, `configupdater` etc. to do the modifications?  For Python files doing modifications consistently will be extremely hard I guess. Would use use RegExes? If an extension would use this feature, even small changes in PyScaffold's templates will most likely lead to strange bugs in the extensions. Sorry for being so pessimistic here, but I really do believe that a powerful feature like this, might lead to a lot of headaches later on.

> * Rely on [`.cirrus.star`](https://cirrus-ci.org/guide/programming-tasks/) and create a shared starlark library to do autodiscovery of some tasks (e.g. just run the type check if a person have a `py.typed` file).
>   
>   * A separated `starlark` lib would also simplify the amount of information a user needs to see for the Cirrus configuration, which would make it less scary to change for starters.
> 
> Of course, those things need to be debated and mainly I need some time to develop them, but I think it could be cool.

Sounds cool. So this would even automize then the switch @CarliJoy mentions. So `.cirrus.yml` would import this lib then and just use its functionality? Sounds good :-) Thanks.

FlorianWilhelm:
> @FlorianWilhelm, do you prefer always performing the typecheck in Cirrus and just failing silently, or use a environment var and `only_if: $TYPE_CHECKING == 'true'` like it is implemented in this PR?
> 
> (Both are basically changing a switch)

Failing silently doesn't sound so good. `only_if: $TYPE_CHECKING == 'true'` sounds better so that a user can switch it on explicitly and if switched on it really fails if something is wrong.

abravalheri:
> Will pyscaffold.file_modifiers use the functionality of other libraries like pyaml, configupdater etc. to do the modifications?

My idea here is to simply define the API in terms of strings to simplify. If the `modifier` function wants to use libraries, it should `parse` the input and `dump` the output... This is more or less what I made in this PR:
https://github.com/pyscaffold/pyscaffold/blob/a332715d6768717c9c2d5781dee2f1192153346d/src/pyscaffold/extensions/typed.py#L102-L118

(and what we are doing in the `update` module)

> For Python files doing modifications consistently will be extremely hard I guess.

I was thinking more in terms of `.cfg` and `.toml` files, since there are libraries that handle it well. This is mainly motivated by the idea of migrating all the configuration to `pyproject.toml`. Each extension would need to add stuff in there...

But yes, I agree that changing Python code is complicated... Nevertheless we are already doing that for `docs/config.py` in `pyscaffoldext-markdown` (and if I am not wrong in `pyscaffold-django`). If someone wants to be on the safe site, they would need to use something like [libCST](https://github.com/Instagram/LibCST) or [refactor](https://refactor.readthedocs.io/en/latest/)

> Sounds cool. So this would even automize then the switch @CarliJoy mentions. So .cirrus.yml would import this lib then and just use its functionality? Sounds good :-) Thanks.

Yes that is the idea, but it is for the future 😝, when I have sometime...
It would simplify a lot the CirrusCI configuration, by abstracting things away.

abravalheri:
Sorry, I messed up the idea between `pyscaffold.file_modifiers` and the idea of having something like `pyscaffold.structure.modify_contents`.

For `pyscaffold.file_modifiers` I was planning something simple, like:
- given 2 regexes (one for start and one for the end)
  - comment everything in between
  - remove everything in between
  - replace everything in between with the new text
- given 1 regex:
  - insert text before
  - insert text after
  - remove line.

It would be more or less a extraction of what we use in `pyscaffoldext-markdown`, `pyscaffold-django`.

FlorianWilhelm (approved):
Thanks, looks good and sorry for the long wait :-)

abravalheri:
Thank you very much for the review

No problems, I will leave this change marinating for a while until I have the time to proper plan for a release 😝 

Maybe rethink somethings...

PR #494, opened 2021-09-10
Update typecheck configs for CI

abravalheri:
This PR attempts to bring changes based on #493 for `tox.ini` and `.cirrus.yml` (this is a long standing tradition of using mostly the same configuration provided by the templates to PyScaffold's own repository and CI).

I have also noticed that, most of the times in Debian docker containers `apt-get install` will fail unless the program is already installed, because `apt-get update` is kind of mandatory (repository caches are probably stripped from the docker images). In the Debian-based Python containers used in Cirrus, git seem to be pre-installed, so there is no need to run that command.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -49,56 +49,54 @@ env:
       - cat .pre-commit-config.yaml src/pyscaffold/templates/pre-commit-config.template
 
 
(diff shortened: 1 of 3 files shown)


FlorianWilhelm:
Nice, thanks!

PR #495, opened 2021-09-10
Mark counter in skeleton explicitly as unused

abravalheri:
Some IDEs might complain that the `i` variable in the for loop of `skeleton.py` is not used. In general, linters accept unused vars when they are explicitly marked with a `_` prefix.

This is also the case of [`flake8-bugbear`](https://pypi.org/project/flake8-bugbear/):

> B007: Loop control variable not used within the loop body.
> If this is intended, start the name with an underscore.

### DIFF
--- a/src/pyscaffold/templates/skeleton.template
+++ b/src/pyscaffold/templates/skeleton.template
@@ -51,7 +51,7 @@ def fib(n):
     """
     assert n > 0
     a, b = 1, 1
-    for i in range(n - 1):
+    for _i in range(n - 1):
         a, b = b, a + b
     return a
 


FlorianWilhelm:
Thanks @abravalheri.

BTW, do you also see in Github in the "Files Changed" tab of a PR a lot of unit test errors even though everyting is green in the "conversation" tab?

abravalheri:
Yes, I do 😝

This happens because, since Python 3.10 finally have a release candidate, I decided to start testing PyScaffold with it, so we would be prepared when the time comes.
Cirrus is configured to run the tests and report the errors, but the build status will not be changed by Python 3.10rc1 results.

PyScaffold seems to be ready for 3.10, but it turns out *Sphinx isn't* (sphinx-doc/sphinx#9505). That is the reason why the integration tests fail on 3.10rc1...

PR #496, opened 2021-09-12
Ensure *--update* includes added extensions in setup.cfg

abravalheri:
Prior to this change, new extensions added with `--update` were not being added to the `[pyscaffold]` section of `setup.cfg`.


FlorianWilhelm:
Oh nice! So that's a long-standing bug fixed then? Cool.

abravalheri:
Thanks @FlorianWilhelm!
Yeah, it is one of those things that I have noticed briefly on the past, but never really paid attention to check if it was actually a bug or just something I was missing...
But working on PyScaffold this weekend I had the chance to investigate it better.

PR #500, opened 2021-09-17
Remove bdist_wheel section from setup.cfg template

### LINKED ISSUE #497: Should we remove the [bdist_wheel] section from setup.cfg template?
Both [packaging guides](https://packaging.python.org/guides/distributing-packages-using-setuptools/#wheels) and [wheels docs](https://wheel.readthedocs.io/en/stable/quickstart.html) seems to suggest that the following section makes sense if the project supports Python 2 and Python 3 at the same time.
```ini
[bdist_wheel]
universal = 1
```

I suppose we don't need this anymore in the template, since it is no longer a "best practice", right?

My understanding is that by default the wheel made is a pure Python3 one, which should work in every system.
(PyScaffold's own `setup.cfg` doesn't have it and things seem to be working fine).

abravalheri:
[Packaging guides] and [wheels docs] seem to suggest that the `bdist_wheel` section is required to support Python 2 and Python 3 at the same time. Since Python 2 was retired, it makes sense to remove it.

PyScaffold's own `setup.cfg` file does not have this section and things work fine (setuptools seem to default to a pure Python 3 package).

Closes #497.

[packaging guides]: https://packaging.python.org/guides/distributing-packages-using-setuptools/#wheels
[wheels docs]: https://wheel.readthedocs.io/en/stable/quickstart.html

### DIFF
--- a/src/pyscaffold/templates/setup_cfg.template
+++ b/src/pyscaffold/templates/setup_cfg.template
@@ -94,10 +94,6 @@ testpaths = tests
 #     slow: mark tests as slow (deselect with '-m "not slow"')
 #     system: mark end-to-end system tests
 
-[bdist_wheel]
-# Use this option if your package is pure-python
-universal = 1
-
 [devpi:upload]
 # Options for the devpi: PyPI server and packaging tool
 # VCS export must be deactivated since we are using setuptools-scm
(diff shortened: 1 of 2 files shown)


FlorianWilhelm:
Thanks @abravalheri!

PR #504, opened 2021-09-26
Avoid 'tool of choice' term in docs

abravalheri:
In the discussion on https://discuss.python.org/t/pyscaffold-joining-cooperating-with-pypa/10789
it was pointed out that people would be more comfortable if we don't use the expression "tool-of-choice", which is a good point.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -34,7 +34,7 @@
 
 |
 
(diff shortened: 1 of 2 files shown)


FlorianWilhelm:
Thanks @abravalheri !

Issue #506, opened 2021-09-29
Update generates invalid config file


CarliJoy:
## Description of the problem
Trying to update a 3.2 project 

**setup.cfg**
```ini
# This file is used to configure your project.
# Read more about the various options under:
# http://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files

[metadata]
name = pyscaffoldext-company
description =  company specific extensions to allow continuous integration
               and best standards within the company
author = Carli Freudenberg
author-email = Carl.Freudenberg@company.de
license = mit
long-description = file: README.rst
long-description-content-type = text/x-rst; charset=UTF-8
url = https://gitlab.company.de/python-devops/pyscaffoldext-company
project-urls =
    Documentation = https://pypi.company.de/company/stable/pyscaffoldext-company/latest/+d/index.html
# Change if running only on Windows, Mac or Linux (comma-separated)
platforms = any
# Add here all kinds of additional classifiers as defined under
# https://pypi.python.org/pypi?%3Aaction=list_classifiers
classifiers =
    Development Status :: 5 - Production/Stable
    Programming Language :: Python :: 3 :: Only
    Topic :: Software Development :: Libraries :: Python Modules
    Topic :: Software Development :: Code Generators
    Operating System :: OS Independent
    Intended Audience :: Developers

[options]
zip_safe = False
packages = find:
include_package_data = True
package_dir =
    =src
# Fix the pyscaffold version so we have fewer things to test
install_requires =
    pyscaffold==3.2.3
    pyscaffoldext-pyproject
    black>=20.8b1
# DON'T CHANGE THE FOLLOWING LINE! IT WILL BE UPDATED BY PYSCAFFOLD!
setup_requires = pyscaffold>=3.2a0,<3.3a0
# Add here dependencies of your project (semicolon/line-separated), e.g.
# install_requires = numpy; scipy
# The usage of test_requires is discouraged, see `Dependency Management` docs
# tests_require = pytest; pytest-cov
# Require a specific Python version, e.g. Python 2.7 or >= 3.4
python_requires = >=3.6

[options.packages.find]
where = src
exclude =
    tests

[options.extras_require]
# Add here additional requirements for extra features, to install with:
# `pip install pyscaffoldext-company[PDF]` like:
# PDF = ReportLab; RXP
# Add here test requirements (semicolon/line-separated)
testing =
    flake8
    pytest
    pytest-cov
    pytest-virtualenv
[...]

PR #508, opened 2021-09-30
Warn user about empty namespaces

abravalheri:
## Purpose
Sometimes extension-related parameters seem to end up not persisted to `setup.cfg`.

Issue #506, shows that it might happen for the `namespace` extension, and via a reaction chain cause errors during updates.

## Approach
The changes implemented in this commit try to prevent this reaction chain of errors happening, by replacing `None` with `""` and instructing the user a workaround for this situation.


FlorianWilhelm:
Thanks @abravalheri!

PR #509, opened 2021-09-30
Prevent multi-line descriptions for crashing putup

abravalheri:
## Purpose

Users can decide to edit `setup.cfg` and change the description field to be a multi-line string.
This can break updates, because the current `setup.cfg` template depends on the fact it is a sinlge line (the INI/CFG syntax is broken otherwise).

This issue is related to #506.

## Approach
When rendering the template use just the first line of the description.
Then use ConfigUpdater to fix it, if necessary, with `set_values([ ... list of lines ...])`.

### DIFF
--- a/src/pyscaffold/templates/__init__.py
+++ b/src/pyscaffold/templates/__init__.py
@@ -135,10 +135,20 @@ def setup_cfg(opts: ScaffoldOpts) -> str:
     Returns:
         str: file content as string
     """
+
     template = get_template("setup_cfg")
-    cfg_str = template.substitute(opts)
+
+    # template assumes oneliner `description`, this workaround avoids parsing errors:
+    desc = opts.get("description", "").splitlines() or [""]
+    cfg_str = template.substitute({**opts, "description": desc[0]})
+
     updater = ConfigUpdater()
     updater.read_string(cfg_str)
+
+    # fix the workaround for multiline description
+    if desc[0] != opts.get("description", ""):
+        updater["metadata"]["description"].set_values(desc)
+
     requirements = deps.add(deps.RUNTIME, opts.get("requirements", []))
     updater["options"]["install_requires"].set_values(requirements)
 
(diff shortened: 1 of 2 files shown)


src/pyscaffold/templates/__init__.py
```diff
@@ -135,10 +135,20 @@ def setup_cfg(opts: ScaffoldOpts) -> str:
     Returns:
         str: file content as string
     """
+
     template = get_template("setup_cfg")
-    cfg_str = template.substitute(opts)
+
+    # template assumes oneliner `description`, this workaround avoids parsing errors:
```
FlorianWilhelm, line 141: Could we write maybe?
# template needs single-line `description`, thus we pass only the first line of a multi-line description...

src/pyscaffold/templates/__init__.py
```diff
@@ -135,10 +135,20 @@ def setup_cfg(opts: ScaffoldOpts) -> str:
...
+    desc = opts.get("description", "").splitlines() or [""]
+    cfg_str = template.substitute({**opts, "description": desc[0]})
+
     updater = ConfigUpdater()
     updater.read_string(cfg_str)
+
+    # fix the workaround for multiline description
```
FlorianWilhelm, line 148:
```
# ... and finally the multi-line string for the full description.
```

FlorianWilhelm:
Thanks. Nice solution. Makes me think if maybe we could one day completely go for ConfigUpdater as template systems for `setup.cfg`. I think it would save us some code and avoid workarounds like this. But I am also not sure if this change would make sense when `pyproject.toml` will replace it anyway.

I added some comments to the code which I think makes it easier to understand what's going on. Feel free to merge @abravalheri. And thanks a lot again :-)

abravalheri:
Yes, that would make sense.
I suppose a purely `ConfigUpdater` does not make sense, because it is just easier to add the structure and the comments in the plain text, but we could avoid doing any string interpolation and just fill the option values using `ConfigUpdater`. It should not be super-hard...

Let's see if we have any news about PEP 621 and setuptools in the near future. If not, we might want to invest sometime on this...

Thank you very much @FlorianWilhelm for the suggestions! I think they improve a lot the PR.

PR #510, opened 2021-09-30
Prevent `None` extension parameters from crashing putup

abravalheri:
## Purpose
As shown in #506, ConfigUpdater will crash when trying to persist `option = None`.
This is problematic when an extension defines a "persitible" parameter that can assume `None`, since it may crash putup.

These changes implement a workaround for that.

## Approach
Force the parameter to be persisted as an empty string instead of None?
This might be controversial, but it is the safest approach for an `INI` format, since `INI` files are opaque in terms of types for option values (in the end of the day they are always strings and it is up for the users to interpret those strings the way they see fit).

## Alternative (not implemented) and Cons:
We could use `allow_no_value=True`, however other tools using the `setup.cfg` (e.g. setuptools, isort, pytest, mypy, etc...), might not like this (the default value for `allow_no_value` in ConfigParser is `False`), and therefore parser errors might crash them.


FlorianWilhelm:
Thanks @abravalheri!

PR #511, opened 2021-09-30
Improve 'debuggability' for update functions

abravalheri:
## Purpose

This PR doesn't change anything about functionalities, but  it is useful for debugging problems like the ones pointed out in #506

## Approach
Just add debug logging messages with the name of the function being executed and the content of the `setup.cfg` file in the main decorator of the `update` module.


FlorianWilhelm:
Thanks @abravalheri!

PR #512, opened 2021-09-30
Raise exception in the case extension is not installed

abravalheri:
## Purpose
There are situations that the user tries to update a previously generated project with PyScaffold, but the extensions are not installed in the current venv.

Currently PyScaffold silently ignores the missing extensions, which can lead to crashes like #506.

## Approach
This PR tries to make this situation obvious and ask for the user to install the missing extensions.

I believe that it is better to panic and raise an exception instead of proceed running PyScaffold, because there are cases that the extensions completely change the desired behaviour (e.g. `namespace` indirectly via `custom_extension`).

### DIFF
--- a/src/pyscaffold/api.py
+++ b/src/pyscaffold/api.py
@@ -7,7 +7,7 @@ from pathlib import Path
 
 from . import __version__ as VERSION
 from . import actions, info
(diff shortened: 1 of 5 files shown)


FlorianWilhelm:
Nice!

Issue #516, opened 2021-10-07
Pyscaffold v4.1.0 cannot handle comments in setup.cfg install_requires section


mhaley-tignis:
## Description of your problem
With pyscaffold v4.0.2 we could have comments in our install_requires section like below

**Please provide a minimal, self-contained, and reproducible example.**
setup.cfg
```python
install_requires =
    importlib-metadata; python_version<"3.8"
#Adding some comments here that are perfectly valid.
    some-other-dependency
```
Running `putup -vv --update test-app` results in the following

**Please provide the full traceback.**
```python
Traceback (most recent call last):
  File "/Users/mhaley/.pyenv/versions/3.8.7/envs/aipc-devtool/lib/python3.8/site-packages/pyscaffold/api.py", line 177, in _read_existing_config
    opts = info.project(opts)
  File "/Users/mhaley/.pyenv/versions/3.8.7/envs/aipc-devtool/lib/python3.8/site-packages/pyscaffold/info.py", line 180, in project
    cfg = read_setupcfg(path, config_file).to_dict()
  File "/Users/mhaley/.pyenv/versions/3.8.7/envs/aipc-devtool/lib/python3.8/site-packages/pyscaffold/info.py", line 256, in read_setupcfg
    updater.read(path, encoding="utf-8")
  File "/Users/mhaley/.pyenv/versions/3.8.7/envs/aipc-devtool/lib/python3.8/site-packages/configupdater/configupdater.py", line 122, in read
    return self._parser().read(filename, encoding, self)
  File "/Users/mhaley/.pyenv/versions/3.8.7/envs/aipc-devtool/lib/python3.8/site-packages/configupdater/parser.py", line 239, in read
    self._read(fp, filename, document)
  File "/Users/mhaley/.pyenv/versions/3.8.7/envs/aipc-devtool/lib/python3.8/site-packages/configupdater/parser.py", line 453, in _read
    self._add_option_line(line)  # HOOK
  File "/Users/mhaley/.pyenv/versions/3.8.7/envs/aipc-devtool/lib/python3.8/site-packages/configupdater/parser.py", line 364, in _add_option_line
    raise InconsistentStateError(msg, self._fpname, self._lineno, line)
configupdater.parser.InconsistentStateError: <Comment: '#Adding some comments here that are perfectly valid.\n'> should be Option or Space
test-app/setup.cfg(51): '    some-other-dependency\n'
```

Changing the section to be like
```
install_requires =
[...]

PR #517, opened 2021-10-07
Handle comments in setup.cfg install_requires section.

mhaley-tignis:
## Purpose
Add support for handing of comments within the install_requirements section in setup.cfg
```
[options]
install_requires =
    importlib-metadata; python_version<"3.8"
    #Adding some comments here that are perfectly valid.
    some-other-dependency
```
https://github.com/pyscaffold/pyscaffold/issues/516

## Approach
_How does this change address the problem?_'

This is a work in progress and is just a test that exercises the case and shows the failure.

## Resources & Links

### DIFF
--- a/src/pyscaffold/info.py
+++ b/src/pyscaffold/info.py
@@ -261,7 +261,7 @@ def read_setupcfg(path: PathLike, filename=SETUP_CFG) -> ConfigUpdater:
     if path.is_dir():
         path = path / (filename or SETUP_CFG)
 
-    updater = ConfigUpdater(inline_comment_prefixes=('#'))
+    updater = ConfigUpdater(inline_comment_prefixes=("#"))
     updater.read(path, encoding="utf-8")
 
     logger.report("read", path)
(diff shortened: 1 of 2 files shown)


abravalheri:
Thank you very much @mhaley-tignis, test cases really help to understand the problem.
Do you want to give the solution go?

I think we can replace all direct the usages of `Requirement(...).name` in `src/pyscaffold/dependencies` with an auxiliary function that wraps it in a try/except...

abravalheri:
Thank you very much @mhaley-tignis, this is a very elegant solution.

There is one drawback though: the removal of the comments in the output (the reason why PyScaffold used ConfigUpdater is to preserve the comments and the original structures of the file).

I need to have a look here if we can use this solution in `read_setupcfg` without compromising the file update, or if it is just easier to fallback and wrap `Requirement(...).name` in a try catch as I previously mentioned...

abravalheri:
Hi @mhaley-tignis, I am closing this one in favour of #518, so we don't throw the comments away...
But I have used your PR as basis, thank you very much!

PR #518, opened 2021-10-09
Fix issue 516 without removing the comments

abravalheri:
## Purpose
This PR uses most of #517 but instead of changing ConfigUpdater syntax options, it prevents `dependencies.deduplicate` from raising syntax errors when trying to interpret comments as PEP 508 specs.

## Approach
A new `pkg_name` function is added to wrap the requirement parsing in a try-except block, returning the original the comment if the parsing is not successful.

### DIFF
--- a/tests/test_update.py
+++ b/tests/test_update.py
@@ -262,6 +262,31 @@ def test_add_dependencies(tmpfolder):
     assert "importlib-metadata" in str(cfg["options"]["install_requires"])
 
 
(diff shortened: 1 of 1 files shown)


abravalheri:
For the time being, I am leaving `attempt_pkg_name`, but we can change later when we have a better name...

I will release another RC for 4.1.1 with all the fixes.

FlorianWilhelm:
Thanks @abravalheri!

PR #521, opened 2021-10-11
Replace pyscaffold name with variable in contributing template

abravalheri:
Users will probably like to call the conda environment by the name of the package...


FlorianWilhelm:
Cool, thanks @abravalheri!

PR #535, opened 2021-10-30
Modernise usage and mentions to `setup.py` in favour of `build`.

abravalheri:
## Purpose
There have been a bunch of changes in packaging practices for Python
since the last time we looked into `test_install`, between them:

- direct invocation of `python setup.py` is now [deprecated]
  (the existence of the file itself does not seem to be deprecated, but running it directly is)
- `setuptools_scm` also deprecated `python setup.py --version`
- `bdist`s, that were already deprecated, should be no longer used by today standards...
- PyPA is actively promoting `python -m build` as main interface

The purpose of this PR is to modernise `test_install` (and hopefully make it faster).

## Approach

- Use `python -m build {--wheel,--sdist}` instead of `python setup.py {sdist,bdist_wheel}`
- Remove `bdist` tests 🎉 
- Remove calls to `python setup.py --version`

I also took the opportunity to use the `venv` module we have in place for the `--venv` extension (reasons: more coverage for that module + I never liked pytest-virtualenv ...).

## Side Effects

- We no longer need to manually manage confusing `tar` files
- Which means that we don't need to install `gnu-tar` in the OS X test on the CI
- Less time creating venvs during tests

[deprecated]: https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -117,9 +117,8 @@ All configuration can be done in ``setup.cfg`` like changing the description,
 URL, classifiers, installation requirements and so on as defined by setuptools_.
 That means in most cases it is not necessary to tamper with ``setup.py``.
 
-In order to build a source or wheel distribution, just run
-``tox -e build`` (``python setup.py sdist`` or ``python setup.py bdist_wheel``
-if you don't use tox_).
+In order to build a source or wheel distribution, just run ``tox -e build``
+(if you don't use tox_, you can also install ``build`` and run ``python -m build``).
 
 .. rubric:: Package and Files Data
 
@@ -147,8 +146,7 @@ the information of tags to infer the version of your project with the help of
 setuptools_scm_.
 To use this feature, you need to tag with the format ``MAJOR.MINOR[.PATCH]``
 , e.g. ``0.0.1`` or ``0.1``.
-Run ``python setup.py --version`` to retrieve the current PEP440_-compliant
-version.
+.. Run ``python -m setuptools_scm`` to retrieve the current PEP440_-compliant version.
 This version will be used when building a package and is also accessible
 through ``my_project.__version__``.
 
--- a/docs/features.rst
+++ b/docs/features.rst
@@ -25,11 +25,9 @@ compliant way, by just running::
     tox -e build
 
 Alternatively, if you are not a huge fan of isolated builds, or prefer running
-the commands yourself, you can execute ``python setup.py bdist_wheel``. Source
-distributions, i.e. ``sdist``, are obsolete and no longer recommended as they
-ignore several options in ``setup.cfg``. If ``universal=1`` in the ``[bdist_wheel]``
-section of ``setup.cfg``, a generated wheel distribution will be as flexible as
-a source distribution and can be installed on every architecture.
+the commands yourself, you can execute ``python -m build --no-isolation .``.
+Source distributions, i.e. ``sdist``, are only recommended if you absolutely need them.
+They are tricky to configure and may ignore several options in ``setup.cfg``.
 
 .. rubric:: Uploading to PyPI
 
@@ -138,7 +136,7 @@ Your project is already an initialised Git repository and setuptools_ uses the
 information of tags to infer the version of your project with the help of
 `setuptools_scm`_.  To use this feature you need to tag with the format
 ``MAJOR.MINOR[.PATCH]`` , e.g. ``0.0.1`` or ``0.1``.
-Run ``python setup.py --version`` to retrieve the current `PEP 440`_-compliant version.
+.. Run ``python -m setuptools_scm`` to retrieve the current `PEP 440`_-compliant version.
 This version will be used when building a package and is also accessible through
 ``my_project.__version__``. If you want to upload to PyPI_ you have to tag the current commit
 before uploading since PyPI_ does not allow local versions, e.g. ``0.0.dev5+gc5da6ad``,
--- a/docs/features.rst
+++ b/docs/features.rst
@@ -25,7 +25,7 @@ compliant way, by just running::
     tox -e build
 
 Alternatively, if you are not a huge fan of isolated builds, or prefer running
-the commands yourself, you can execute ``python -m build --no-isolation .``.
+the commands yourself, you can execute ``python -m build --wheel --no-isolation``.
 Source distributions, i.e. ``sdist``, are only recommended if you absolutely need them.
 They are tricky to configure and may ignore several options in ``setup.cfg``.
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -76,11 +76,15 @@ ds =
     pyscaffoldext-dsproject>=0.5
 # Add here test dependencies (used by tox)
 testing =
+    # Used for building/installing during tests
     setuptools
+    setuptools_scm[toml]
+    wheel
+    build
+    #
     tomlkit     # as dependency in `-e fast`
     certifi     # tries to prevent certificate problems on windows
     tox         # system tests use tox inside tox
-    build       # system tests use it to build projects
     pre-commit  # system tests run pre-commit
     sphinx      # system tests build docs
     flake8      # system tests run flake8
@@ -122,7 +126,7 @@ addopts =
     --verbose
 #    In order to use xdist, the developer can add, for example, the following
 #    arguments:
-#    --dist=load --numprocesses=auto
+#    --numprocesses=auto
 norecursedirs =
     dist
     build
--- a/src/pyscaffold/shell.py
+++ b/src/pyscaffold/shell.py
@@ -16,6 +16,8 @@ from .log import logger
 
 PathLike = Union[str, os.PathLike]
 
+IS_POSIX = os.name == "posix"
+
 # The following default flags were borrowed from Github's docs:
 # https://docs.github.com/en/github/getting-started-with-github/associating-text-editors-with-git
 EDITORS: Dict[str, List[str]] = {
@@ -86,8 +88,11 @@ class ShellCommand:
             "universal_newlines": True,
             **kwargs,  # allow overwriting defaults
         }
-        return subprocess.run(command, **opts)
-        # ^ `check_output` does not seem to support terminal editors
+        if self._shell:
+            return subprocess.run(command, **opts)
+            # ^ `check_output` does not seem to support terminal editors
+        return subprocess.run(shlex.split(command, posix=IS_POSIX), **opts)
+        # ^ Joining and then splitting is the only way of supporting all the cases
 
     def __call__(self, *args, **kwargs) -> Iterator[str]:
         """Execute the command, returning an iterator for the resulting text output"""
@@ -242,3 +247,4 @@ def join(parts: Iterable[Union[str, PathLike]]) -> str:
 
 #: Command for git
 git = get_git_cmd()
+python = ShellCommand(sys.executable)
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -112,7 +112,7 @@ linux_mac_task:
       brew_cache:
         folder: "$HOME/Library/Caches/Homebrew"
       install_script:
-        - brew install python gnu-tar
+        - brew install python
         - brew cleanup
   <<: *REGULAR_TASK_TEMPLATE
 
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -29,4 +29,5 @@ jobs:
         TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
       run: |
         python -m pip install --upgrade pip setuptools tox
-        python -m tox -e clean,build,publish -- --verbose --repository pypi
+        python -m tox -e clean,build
+        python -m tox publish -- --verbose --repository pypi
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -291,8 +291,7 @@ you are trying to run the CI scripts in a fork repository, make sure to push
 all the tags.
 You can also try to remove all the egg files or the complete egg folder, i.e.,
 ``.eggs``, as well as the ``*.egg-info`` folders in the ``src`` folder or
-potentially in the root of your project. Afterwards run ``python setup.py
-egg_info --egg-base .`` again.
+potentially in the root of your project.
 
 ..
 
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -304,7 +304,7 @@ Best Practices and Common Errors with Version Numbers
 How do I get a clean version like 3.2.4 when I have 3.2.3.post0.dev9+g6817bd7?
     Just commit all your changes and create a new tag using ``git tag v3.2.4``.
     In order to build an old version checkout an old tag, e.g. ``git checkout -b v3.2.3 v3.2.3``
-    and run ``tox -e build``.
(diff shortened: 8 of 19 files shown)


src/pyscaffold/shell.py
```diff
@@ -86,8 +88,11 @@ def run(self, *args, **kwargs) -> subprocess.CompletedProcess:
             "universal_newlines": True,
             **kwargs,  # allow overwriting defaults
         }
-        return subprocess.run(command, **opts)
-        # ^ `check_output` does not seem to support terminal editors
+        if self._shell:
+            return subprocess.run(command, **opts)
+            # ^ `check_output` does not seem to support terminal editors
+        return subprocess.run(shlex.split(command, posix=IS_POSIX), **opts)
```
abravalheri, line 94: This might have been an untested error... I was facing very cryptic errors.

I think originally the class was not thought as receiving a multi-word command string such as `ShellCommand("code --wait")`, but then we changed that for the EDITOR env variable... That might have been the point where things became more complicated.

abravalheri:
https://github.com/pyscaffold/pyscaffold/blob/1eb26ba921b4f7f77c658f8354b250cc4839ec3c/CONTRIBUTING.rst#L292-L295

Do we still need this instruction now that PyScaffold no longer `setuptools` and therefore does not need to bootstrap itself?

BTW, do we still need to build `sdists`?

abravalheri:
https://github.com/pyscaffold/pyscaffold/blob/1eb26ba921b4f7f77c658f8354b250cc4839ec3c/src/pyscaffold/templates/tox_ini.template#L34-L37

Since our docs say `sdist`s are deprecated, should we add the `--wheel` flag by default here?

docs/features.rst
```diff
@@ -25,11 +25,9 @@ compliant way, by just running::
...
 
 Alternatively, if you are not a huge fan of isolated builds, or prefer running
-the commands yourself, you can execute ``python setup.py bdist_wheel``. Source
-distributions, i.e. ``sdist``, are obsolete and no longer recommended as they
-ignore several options in ``setup.cfg``. If ``universal=1`` in the ``[bdist_wheel]``
-section of ``setup.cfg``, a generated wheel distribution will be as flexible as
-a source distribution and can be installed on every architecture.
+the commands yourself, you can execute ``python -m build --no-isolation .``.
+Source distributions, i.e. ``sdist``, are only recommended if you absolutely need them.
+They are tricky to configure and may ignore several options in ``setup.cfg``.
```
abravalheri, line 30: Should we keep saying `sdist` are deprecated? `bdist` definetely are, but the `build` package and PEP517 seem to be fine with `sdist` (or at least do not say officially they are deprecated).

Regardless, since `sdist`s are at least tricky, should we change the `tox.ini` template to not build them by default?

docs/features.rst
```diff
@@ -138,7 +136,7 @@ Your project is already an initialised Git repository and setuptools_ uses the
 information of tags to infer the version of your project with the help of
 `setuptools_scm`_.  To use this feature you need to tag with the format
 ``MAJOR.MINOR[.PATCH]`` , e.g. ``0.0.1`` or ``0.1``.
-Run ``python setup.py --version`` to retrieve the current `PEP 440`_-compliant version.
+.. Run ``python -m setuptools_scm`` to retrieve the current `PEP 440`_-compliant version.
```
abravalheri, line 139: This is not officially supported yet. I recently submitted a PR for that.

setup.cfg
```diff
@@ -122,7 +126,7 @@ addopts =
     --verbose
 #    In order to use xdist, the developer can add, for example, the following
 #    arguments:
-#    --dist=load --numprocesses=auto
```
abravalheri, line 125: I can no longer find docs for this option.

abravalheri:
There are other tests that could benefit from some modernisation as well (I guess `test_update` and system tests), but that is left for the future :P

abravalheri:
It might be just impression, but I think the tests are running faster:

image

The PR remove some of them, and for each test removed there was a creation of a virtual env, so at least in theory, it should be faster 😅 (but maybe it is not significant in the hardware used by Cirrus...)

README.rst
```diff
@@ -117,9 +117,8 @@ All configuration can be done in ``setup.cfg`` like changing the description,
 URL, classifiers, installation requirements and so on as defined by setuptools_.
 That means in most cases it is not necessary to tamper with ``setup.py``.
 
-In order to build a source or wheel distribution, just run
-``tox -e build`` (``python setup.py sdist`` or ``python setup.py bdist_wheel``
-if you don't use tox_).
+In order to build a source or wheel distribution, just run ``tox -e build``
+(if you don't use tox_, you can also install ``build`` and run ``python -m build``).
```
FlorianWilhelm, line 121: Really good! I was actually asked a few times if one needs ``tox`` to build, thus I think mentioning the ``build`` package here is really important. Maybe it could also be mentioned additionally in the parts of the docs where we explain how to build a package.

docs/features.rst
FlorianWilhelm: Great suggestion. I think the default should be wheel-only. I see no reason why one needs source dists and it's good that we tell at that point that they are tricky and ignore options.

docs/features.rst
FlorianWilhelm: Nice! Didn't know that :-) I would love to see this happen. In general, we would come to a point where our docs no longer mention ``python setup.py ...`` while still providing the same functionality.

setup.cfg
FlorianWilhelm: Thanks for checking. It's good to remove clutter like this then.

src/pyscaffold/shell.py
FlorianWilhelm: It's funny that after all those years, there are still problems coming up due to calling shell commands. Calling external commands like this is always a minefield and should thus be avoided, at least that's what I keep telling our junior developers. In our case sadly there is no other way as Git has no real library implementation :-(

FlorianWilhelm (approved):
Thanks. I really enjoyed reviewing this PR. I feel like the code got a lot better and was easier to read.

docs/features.rst
abravalheri: Should we also only release wheels for PyScaffold itself in the future? (So I can apply the same changes for `tox.ini` and `tox_ini.template`? 😝

src/pyscaffold/shell.py
abravalheri: Hopefully it works fine now!
(I think this internal inconsistency was just introduced accidentally when we started accepting "multi-word" commands for the `$EDITOR`, before that I am sure it was fine...)

abravalheri:

I added some other changes:
- Mention the `build` package a bit more in the docs as an alternative to `tox -e build`
- Changed `tox.ini` and `tox_ini.template` to build only wheels by default.

This also means that the next PyScaffold release will be wheel-only. Is that fine for you?

docs/features.rst
FlorianWilhelm: Yes, that makes sense! We needed this only until 3.x since setuptools was a development requirement until then, i.e. ``setup_requires``. As this is no longer needed we should drop it and also remove all the notes about ``python setup.py egg_info --egg-base .`` as you mentioned above.

src/pyscaffold/shell.py
FlorianWilhelm: Thanks :-)

FlorianWilhelm:
Thanks @abravalheri! Looks really good to me.

abravalheri:
According to [pypa/setuptools#2835](https://github.com/pypa/setuptools/pull/2835#issuecomment-955821548), creating a `sdist` is pretty much part of the process of creating a `wheel` (there is evidence for that too when checking the logs for `python -m build`).

On the light of the new events, there is no much point on saying that users should use them only when necessary :( (which means I should revert some of the changes).

Still, there is a [bug](https://github.com/pypa/setuptools/issues/1461#issuecomment-749092366) that makes the usage of `include_package_data=True` and `package_data` complicated.

The original text mention:

> Source distributions, i.e. ``sdist``, are obsolete and no longer recommended as they ignore several options in ``setup.cfg``.

@FlorianWilhelm, do you know if there other options being ignored other than the `(include_|exclude_)?package_data` ones?

FlorianWilhelm:
Hm... I am not sure I got the whole picture here. So if building a `wheel` means first building a `sdist` followed by some more magic, why can't we still state that a pure `sdist` format (so without the wheel step) is not recommended? I would not even mention this complicated detail that `sdist` is a step in creating a `wheel`. From a package publisher perspective, all one needs to know is that `wheel` files work as expected and even respect `include_package_data`. 

Regarding your question. No, I don't know if other options are ignored. In my last experiments, which I did a long time ago, I always had to use `MANIFEST.in` in order to add non-python files to my `sdist` package and this I found always not user-friendly. It's just another strange config file.

abravalheri:
Thanks @FlorianWilhelm

Ok, let's keep the recommendation of creating wheels only. 

>  I always had to use MANIFEST.in in order to add non-python files to my sdist package and this I found always not user-friendly.

Yeap, that is the bug I mentioned that happens when `include_package_data=True`. Hopefully they will fix it someday.

>  It's just another strange config file.

If I understood it right, the `MANIFEST.in` file is just something `setuptools` inherited from an era when git/other SCMs were not that popular... What it is actually doing is something similar to `.gitignore` (avoiding unwanted files to get packaged).

I think that makes sense for `setuptools`... they have to be generic and therefore account for these situations.

FlorianWilhelm:
> > I always had to use MANIFEST.in in order to add non-python files to my sdist package and this I found always not user-friendly.
> 
> Yeap, that is the bug I mentioned that happens when `include_package_data=True`. Hopefully, they will fix it someday.

But according to the actual [bug report](https://github.com/pypa/setuptools/issues/1461#issue-351795819) `include_package_data=True` means that only `MANIFEST.in` is used not `package_data`, but in combination with git and `setuptools_scm` no `MANIFEST.in` is needed as also mentioned [here](https://setuptools.pypa.io/en/latest/userguide/datafiles.html). This is really, really confusing. 

> If I understood it right, the MANIFEST.in file is just something setuptools inherited from an era when git/other SCMs were not that popular... What it is actually doing is something similar to .gitignore (avoiding unwanted files to get packaged).

Yes, but wouldn't it make sense to get rid of `MANIFEST.in` and to use some keywords within `setup.cfg` to specify this? One could for instance have `include_scm_pkg_data=True`, which does what `include_package_data=True` does in combination with `setuptool_scm` so include non-Python files within the package that are tracked by an SCM. With `include_pkg_data` and `exclude_pkg_data` one could further refine this. For files outside a package, one could have the *not-recommended* `include_nonpkg_data` and `exclude_nonpkg_data` or something like this.  I see many use-cases for the latter two. The LICENSE file should always be shipped and for the docs one could have `include_docs`. Maybe I am oversimplifying this but in that specific case, setuptools is quite confusing and not consistent in its naming.

abravalheri:
Sorry for bringing this complicated topics up, I completely get how confusing/frustrating it is 😅 

I would argue that the bug report is incomplete. In fact I conducted some [investigation](https://github.com/abravalheri/experiment-setuptools-package-data) myself over the weekend, with the assumption that the data wouldn't be subject to human errors (as docs do) 😝

This is a bottom-up conclusion in terms of boolean algebra (obtained directly from the data via [Quine-McCluskey algorithm](https://en.wikipedia.org/wiki/Quine%E2%80%93McCluskey_algorithm) - sorry for the clumsy notation):

- _i_: ``include_package_data=True``
- _e_: ``exclude_package_data`` is set to remove "data files"
- _p_: ``package_data`` is set to include "data files"
- _m_: ``MANIFEST.in`` exists and includes the "data files"
- _w_: "data files" included in the wheel distribution
- _s_: "data files" included in the sdist distribution

- _!x_: the negated form of ``x`` - ``!x = not(x)``
- _x.y_: logical ``and`` between ``x`` and ``y``
- _x+y_: logical ``or`` between ``x`` and ``y``

```haskell
 s = !i.!e.p + m    (1) 
 w = i.!e.m + !e.p  (2)
```

Retrofitting [jaraco's comment about sdist being a step to produce wheel](https://github.com/pypa/setuptools/pull/2835#issuecomment-955821548), we can prove that the following expression (3) is indeed equivalent to (2) and therefore demonstrated by the experiment:

```haskell
w = i.!e.s + !e.p  (3)
```

**The bug here is that _``!i``_ in (1)** (confirmed by [jaraco's comment about *"hacky behaviour"*](https://github.com/pypa/setuptools/issues/1461#issuecomment-749092366)). If it wasn't for it we would have:

```haskell
s = !e.p + m 
w = !e.p + i.!e.s
```

Which would mean that `MANIFEST.in` is completely unnecessary, and we could only rely on `setuptools_scm` and `package_data` regardless of the type of the distribution.

According to my summary in https://github.com/pypa/setuptools/pull/2835#issuecomment-956123517, I understand that the **_desired_** behaviour is:


1. default setuptools algorithm/scm plugins/``MANIFEST.in`` list files to be included in the `sdist`
2. ``include_package_data=True`` allow all the files inside the package directories in the `sdist` to be added to the `wheel`.
3. ``package_data`` **_SHOULD_** add extra files (e.g. automatically generated files) to both ``sdist`` and ``wheel``.
4. ``exclude_package_data`` refines the behaviour of ``include_package_data`` and ``package_data``.

Which would not be that bad... (still all the problems with names and non-ideal defaults you mention would exists)

---
I haven't added `setuptools_scm` to the mix, so that is something I can improve in the experiment in the future. In theory it should have the same effect as `m`, otherwise it is a bug I would have to report.

FlorianWilhelm:
Never seen this approach to a bug, you are quite a scientist :-) I must say that I really like this approach as you found the actual culprit and found out the connection between `sdist` and `wheel`. Does this now allow for a minimal-invasive bugfix in `setuptools`?

Maybe one stupid question. You say:
> 2. include_package_data=True allow all the files inside the package directories in the sdist to be added to the wheel.

Wouldn't this mean in your [results table](https://github.com/abravalheri/experiment-setuptools-package-data#results) that if `include_package_data=True` **and** `sdist =True` (so `.txt` is in it) it should follow `wheel=True`? This is not case if `exclude_package_data=Yes` so the desired behavior is not even close to the actual behavior, right?

Do you think the `setuptools` maintainers will fix this behavior as it would affect a lot of packages? A workaround could be to leave the current behavior as it is and to introduce new keywords with the desired behavior. Mixing both families of keywords should result in an exception. This would allow for a smooth transition over many `setuptools` versions until the old family of keywords is deprecated and removed.

TIL: Quine–McCluskey algorithm for minimizing boolean functions, not bad for a Monday ;-)

abravalheri:
> So if in row 11 of your [results table](https://github.com/abravalheri/experiment-setuptools-package-data#results) we would have `sdist=True` as in row 3 everything would be fine, right? 

I think so 😄 

Regarding PRs and fixing setuptools, I am preparing a [PR](https://github.com/pypa/setuptools/pull/2844), let's see what is the feedback from the maintainers. My personal opinion is that it is a bit of a stretch if someone is relying in this buggy behaviour somehow...

As you noticed, it should only affect 1 situation of that table, and the outcome would be making the files inside the `sdist` equivalent to the `wheel`. I cannot think in a situation where someone wants purposefully to have less files in a `sdist` than in the equivalent `wheel`...

PR #541, opened 2021-11-22
Ignore specific Pipenv error on Python 3.6

abravalheri:
## Purpose
It seems that the latest version of Pipenv breaks the tests for Python 3.6

## Approach
I don't think the problem is related to PyScaffold itself, so I am inclined to just ignore it.
(Moreover, Python 3.6 EoS is coming this year).

So it should be fair to just skip this test on Python 3.6...

### DIFF
--- a/tests/system/test_dependency_managers.py
+++ b/tests/system/test_dependency_managers.py
@@ -1,5 +1,6 @@
 import json
 import os
+import sys
 from functools import partial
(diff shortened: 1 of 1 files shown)


FlorianWilhelm:
Cool, let's just sit this out, best solution here. Thanks @abravalheri.

Issue #540, opened 2021-11-16
Second header of Section 6.3 in the documentation PDF doesn't wrap


arden13:
## Description of your problem

On the [latest stable documentation pdf](https://readthedocs.org/projects/pyscaffold/downloads/pdf/stable/) at page 40 section the second header doesn't wrap for some reason. I think it just needs re-rendered. A crude screenshot is below.

![image](https://user-images.githubusercontent.com/21027568/142025146-d1a85666-fecd-422e-9cdc-26defe46078f.png)

abravalheri:
Thank you very much @arden13 . This is probably related to the way Latex renders definition lists. Maybe we have to use titles in the rst after all.

I will keep it in mind when I have some time, but please feel free to fire some PRs too ❤

abravalheri:
Hi @arden13, I think @FlorianWilhelm have kindly fixed this problem in his PR, so I will close this issue for now. But please feel free to open it again if you think the are things missing.

arden13:
> Hi @arden13, I think @FlorianWilhelm have kindly fixed this problem in his PR, so I will close this issue for now. But please feel free to open it again if you think the are things missing.

Looks great, I just checked the latest version. Thanks all!

PR #543, opened 2021-11-28
Fix long headline in PDF

FlorianWilhelm:
This fixes issue #540 by reformulating the question in the FAQ. Does it still have the same meaning @abravalheri?

Also, the copyright was changed to PyScaffold Contributors.


abravalheri:
Thank you very much @FlorianWilhelm. I think the change work very well.

Issue #544, opened 2021-11-30
Warn the user or error if a project scaffold is nested


FlorianWilhelm:
## Description of your problem

**Please provide a minimal, self-contained, and reproducible example.**
```bash
putup outer_proj
cd outer_proj
putup inner_proj
cd inner_proj
python setup.py --version
```

**Please provide the full traceback.**
```bash
LookupError: setuptools-scm was unable to detect version for '/Users/fwilhelm/Sources/outer_proj/inner_proj'.

Make sure you're either building from a fully intact git repository or PyPI tarballs. Most other sources (such as GitHub's tarballs, a git checkout without the .git folder) don't contain the necessary metadata and will not work.
```

**Please provide any additional information below.**

The problem is that git detects the git repository of `outer_proj` and doesn't create a new repository at the root of `inner_proj`, which is the intended way for git. The problem is now that `setuptool_scm` needs to be directed to the parent repository to actually retrieve the necessary details. So it fails like shown above. 
One could configure the parent directory for `setuptool_scm` but this would also surely not match the expectation of the user. My suggestion would thus be to check at the invocation of `putup` that the target directory isn't already a repository. If it is we should fail and tell the user about it. I don't think this is intended anyway and only leads to head scratches when you get this mysterious error message of `setuptools_scm`. 

## Versions and main components

* PyScaffold Version: 4.1.1.post1.dev28+g075b76f
* Python Version: 3.8
* Operating system: Linux
* How did you install PyScaffold: (conda/pip) pip (-e)

PR #545, opened 2021-11-30
Prevent accidental creation of nested projects

abravalheri:
## Purpose
This PR attempts to prevent users from accidentally creating nested projects, which may result in problems with `setuptools-scm`, as described in #544.

## Approach
The `pyscaffold.actions.verify_project_dir` was modified to also look for nested repositories and raise an error when relevant.

If the `--force` option is given, PyScaffold will understand that the users know what they are doing and create the project without complaining.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,12 +3,14 @@ Changelog
 =========
 
 
-..
-    Development version
-    ===================
+Development version
+===================
 
-    Version 4.X.X, 2021-XX-XX
-    -------------------------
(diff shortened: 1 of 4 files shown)


abravalheri:
@FlorianWilhelm, is this more or less what you had in mind?

FlorianWilhelm:
Wow! You are so fast 🚀   Looks really good :-) 

BTW, I noticed yesterday again how fast the unit tests are. Thanks again for having sped them up so much ⚡

Issue #551, opened 2021-12-16
Missing flag in the publish-package GitHub workflow


git-afsantos:
## Description of your problem

As per the discussion in https://github.com/pyscaffold/pyscaffold/discussions/550:

> The last step of the `publish-package.yml` workflow runs the following commands to build and publish the package to PyPI.
> 
> ```shell
> python -m pip install --upgrade pip setuptools tox
> python -m tox -e clean,build
> python -m tox publish -- --verbose --repository pypi
> ```
> 
> The [last line](https://github.com/pyscaffold/pyscaffold/blob/de7aa5dc059fbd04307419c667cc4961bc9df4b8/.github/workflows/publish-package.yml#L33) in particular, seems to be missing the `-e` option. Without it, `tox` would run the default environment, which, in this case, runs `pytest`.
> 
> It seems however that tox managed to guess and do the right thing (I can see the latest releases were uploaded to PyPI). But this behaviour is undocumented and should not be relied upon...

## Versions and main components

* PyScaffold Version: v4.1.1
* Python Version: any
* Operating system: any
* How did you install PyScaffold: pip

git-afsantos:
I will submit a PR to fix this issue shortly.

abravalheri:
Thank you very much for the help André!

PR #552, opened 2021-12-16
Add missing -e flag to the publish-package GitHub workflow

git-afsantos:
## Purpose

This Pull Request adds the missing `-e` option to the `tox -e publish` command. It fixes the issue reported in [pyscaffold#551](https://github.com/pyscaffold/pyscaffold/issues/551).

It is a straightforward fix that should not affect anything else, so I believe further details are unnecessary.


abravalheri:
Thank you very much André!!
Not only for identifying the problem but also submitting a fix.

PR #553, opened 2021-12-18
Fix typecheck error 'Incompatible types in assignment'

FlorianWilhelm:
This fixes:
```
src/pyscaffold/log.py: note: In member "wrapped" of class "ReportLogger":
src/pyscaffold/log.py:221: error: Incompatible types in assignment (expression
has type "Optional[Any]", variable has type "Handler")  [assignment]
            self.handler = getattr(self, "_handler", None)
                           ^
Found 1 error in 1 file (checked 29 source files)
```

### DIFF
--- a/src/pyscaffold/log.py
+++ b/src/pyscaffold/log.py
@@ -4,10 +4,10 @@ Custom logging infrastructure to provide execution information for the user.
 import logging
 from collections import defaultdict
 from contextlib import contextmanager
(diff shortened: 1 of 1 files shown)


abravalheri:
Thank you very much @FlorianWilhelm .

I am just noticing how these setters and getters seem convoluted now. Hopefully it can be simplified when we move to rich.

PR #554, opened 2021-12-20
[CI] Update git setup for Windows in CirrusCI

abravalheri:
Maintenance PR that updates the version of GitForWindows in the CirrusCI setup (and also the version of nuget used as installer)


FlorianWilhelm:
Thanks! :-)

Issue #556, opened 2021-12-29
subprocess.run('git --version')  returns '& was unexpected at this time.\n'  and false error of git not installed


ksteptoe:
## Description of your problem
running putup returns ...
ERROR: Make sure git is installed and working.

GIt is installed and working both in powershell and Git Bash. 
Tested using git --version 
PATH variables are set and checked.

I made a study of the issue and proved that git actually runs
but that the subprocess.run method returns an error snf 
stdout returns :
'& was unexpected at this time.'

**Please provide a minimal, self-contained, and reproducible example.**
The following snippet  is provided from the method run in 'ShellCommand class'

```python
#        return subprocess.run(command, **opts)
        # ^ `check_output` does not seem to support terminal editors

    def __call__(self, *args, **kwargs) -> Iterator[str]:
        """Execute the command, returning an iterator for the resulting text output"""
        completed = self.run(*args, **kwargs)
        try:
            completed.check_returncode()
        except subprocess.CalledProcessError as ex:
            msg = "\n".join(e or "" for e in (completed.stdout, completed.stderr))
            raise ShellCommandException(msg) from ex
```
**Debugging,  I find that the returned completed object is returned as follows:**
* CompletedProcess(args='git --version', returncode=1, stdout='& was unexpected at this time.\n')*

**Please provide the full traceback.**
```python
# Traceback (most recent call last):
  File "C:\Users\kevin\miniconda3\lib\site-packages\pyscaffold\exceptions.py", line 33, in func_wrapper
    func(*args, **kwargs)
  File "C:\Users\kevin\miniconda3\lib\site-packages\pyscaffold\cli.py", line 263, in run
    main(args or sys.argv[1:])
  File "C:\Users\kevin\miniconda3\lib\site-packages\pyscaffold\cli.py", line 256, in main
    opts["command"](opts)
  File "C:\Users\kevin\miniconda3\lib\site-packages\pyscaffold\cli.py", line 225, in run_scaffold
    api.create_project(opts)
  File "C:\Users\kevin\miniconda3\lib\site-packages\pyscaffold\api.py", line 156, in create_project
    return reduce(actions.invoke, pipeline, ({}, opts))
  File "C:\Users\kevin\miniconda3\lib\site-packages\pyscaffold\actions.py", line 98, in invoke
    return action(*struct_and_opts)
  File "C:\Users\kevin\miniconda3\lib\site-packages\pyscaffold\actions.py", line 221, in get_default_options
    info.check_git()
[...]

PR #559, opened 2021-12-30
Always log shell ops

FlorianWilhelm:
In issue #556 came up the problem that finding the real reason behind a `ERROR: Make sure git is installed and working.`-error was not easy to figure out using --very-verbose.

This PR now always logs shell commands. This actually reduces code complexity as we always log and let the log level decide what we show. This behaviour for ShellCommand is then more consistent with respect to how the rest of PyScaffold works.


FlorianWilhelm:
Okay, so tests pass only the coverage was reduced due to the fact that code was eliminated. That's fine.

abravalheri (approved):
👏

Thank you very much @FlorianWilhelm, you even managed to have a bigger amount of removed lines than added ones!

PR #560, opened 2021-12-30
Fix problem with Win10 shell

FlorianWilhelm:
This fixes the problem with the Windows shell in issue #556. The basic solution is to add some more environment variables as stolen from setuptools-scm's `utils.do_ex` and in case of Windows try to find out if it is better to call a command with `shell=True` or `shell=False`. It seems that this differs from time to time.

**Note**: Only review and merge after having merged PR #559 into master as I needed to rebase it on #559 so the diff is actual way smaller.

### DIFF
--- a/src/pyscaffold/shell.py
+++ b/src/pyscaffold/shell.py
@@ -177,7 +177,7 @@ def get_git_cmd(**args):
     if IS_WINDOWS:  # pragma: no cover
         # ^  CI setup does not aggregate Windows coverage
         for shell in (True, False):
-            git = ShellCommand("git", shell=shell, env=env, **args)
+            git = ShellCommand("git.exe", shell=shell, env=env, **args)
             try:
                 git("--version")
             except ShellCommandException:
--- a/src/pyscaffold/shell.py
+++ b/src/pyscaffold/shell.py
@@ -104,13 +104,13 @@ class ShellCommand:
         completed = self.run(*args, **kwargs)
         try:
             completed.check_returncode()
-        except subprocess.CalledProcessError as ex:
+        except (subprocess.CalledProcessError, FileNotFoundError) as e:
             stdout, stderr = (e or "" for e in (completed.stdout, completed.stderr))
             stdout, stderr = (e.strip() for e in (stdout, stderr))
             sep = " :: " if stdout and stderr else ""
             msg = sep.join([stdout, stderr])
-            logger.debug(f'last command failed with "{msg}"')
-            raise ShellCommandException(msg) from ex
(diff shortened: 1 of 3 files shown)


abravalheri:
Just for my own tracking, this URL should give us the diff between the 2 branches instead of the comparison with master:

https://github.com/pyscaffold/pyscaffold/compare/fix-556-debugging...fix-556-win-shellcommand

src/pyscaffold/shell.py
```diff
@@ -132,10 +149,10 @@ def get_git_cmd(**args):
     Args:
         **args: additional keyword arguments to :obj:`~.ShellCommand`
     """
-    if sys.platform == "win32":  # pragma: no cover
+    if IS_WINDOWS:  # pragma: no cover
         # ^  CI setup does not aggregate Windows coverage
-        for cmd in ["git.cmd", "git.exe"]:
```
abravalheri, line 137: I understand from this removal that `"git.cmd"` is no longer necessary, and that we expect any distribution of git for Windows to add a `.exe` to the PATH, right?

src/pyscaffold/shell.py
```diff
@@ -132,10 +149,10 @@ def get_git_cmd(**args):
...
-            git = ShellCommand(cmd, **args)
+        for shell in (True, False):
+            git = ShellCommand("git.exe", shell=shell, **args)
```
abravalheri, line 155: I am very curious to understand what is happening for the command with `shell=True` to be failing...

Would be a problem with file names?

abravalheri (approved):
I rebased the PR, and it should be ready to merge once the CI passes.

I am just curious about some behaviours

abravalheri:
Thank you very much @FlorianWilhelm for debugging and solving this problem even without access to a Windows environment. That was epic!

❤️ 💜 💚 💛 💙

src/pyscaffold/shell.py
FlorianWilhelm: Yes. I was actually trying to find `git.cmd` by googling but had no luck. There is a "git CMD" but no "git.cmd" so maybe it was a mistake? Also setuptools-scm doesn't seem to be doing this, so at least we would fail both.

src/pyscaffold/shell.py
FlorianWilhelm: Yea, I am also very curious about this. We only now that one of @ksteptoe's two Windows machines seem to have a problem with the shell setup. Somehow an `&` is added but I have no good explanation for that. I think it will be impossible to figure this out without access to the machine. Also I am no Windows (Power)shell expert.

PR #564, opened 2022-01-10
Improve off-by-one error handling in test_install

### LINKED ISSUE #563: setuptools-scm seems to break unit tests
Our [unittests break ](https://cirrus-ci.com/task/4762004492648448) in `test_sdist_install_dirty` and `test_sdist_install_with_1_0_tag_dirty`. Strangely those tests fail only under `test (OS X)` and `coverage (Linux - 3.6)`.

This is the stacktrace.
```python
AssertionError: assert (False or False)
  +  where False = <built-in method endswith of str object at 0x103224300>('.d20220109')
  +    where <built-in method endswith of str object at 0x103224300> = 'ge9df8ac.d20220110'.endswith
  +  and   False = <built-in method endswith of str object at 0x103224440>('.d2022010')
  +    where <built-in method endswith of str object at 0x103224440> = 'ge9df8ac.d2022011'.endswith
```
The stacktrace shows that somehow the new automatic generated version string is the date like `d20220109`. This is quite strange as we use `no-guess-dev` for setuptools-scm's configuration. Strange is that the current setuptools-scm version is `6.3.2` from September 21th, 2021. So why does this appear now?

* PyScaffold Version: master
* Python Version: *
* Operating system: Mac/Linux
* How did you install PyScaffold: (conda/pip)

abravalheri:
## Purpose
It seems that `test_install` can be flaky if running around midnight or
there is some problem with the time configuration in Python/Git.

The previous workaround worked for most situations, but not always.
For example, it could fail when the test was running at midnight of the
day 09 => 10, because these numbers differ in the first digit.

This implementation (hopefully) improves this and fix #563.

## Approach
Instead of comparing the date without the last digit, do extra comparisons with the date +/-1.

### DIFF
--- a/tests/test_install.py
+++ b/tests/test_install.py
@@ -13,9 +13,9 @@ can execute in parallel and not interfere with each other.
 import os
 import shutil
 from contextlib import contextmanager
+from datetime import datetime, timezone
 from pathlib import Path
 from shutil import copyfile
-from time import strftime
 
 import pytest
 
@@ -183,24 +183,22 @@ class DemoApp:
 
 
 def check_version(version, exp_version, dirty=False):
-    today = int(strftime("%Y%m%d"))
-    yesterday, tomorrow = today - 1, today + 1
-    # ^  sometimes the day in the dirty tag has a 1-off error ¯\_(ツ)_/¯
-    #    specially in the case the test is running around midnight
-    dirty_tags = [f".d{d}" for d in (today, yesterday, tomorrow)]
+    today = datetime.now(timezone.utc).strftime("%Y%m%d")
+    dirty_tag = f".d{today}"
     # ^  this depends on the local strategy configured for setuptools_scm...
     #    the default 'node-and-date'
 
-    # for some setuptools version a directory with + is generated, sometimes _
     sep = "+" if "+" in version else "_"
+    # ^  for some setuptools version a directory with + is generated, sometimes _
+
     if dirty:
         ver, local = version.split(sep)
-        assert any(local.endswith(d) for d in dirty_tags)
+        assert local.endswith(dirty_tag)
         assert ver == exp_version
     else:
         ver = version.split(sep)
         if len(ver) > 1:
-            assert not any(ver[1].endswith(d) for d in dirty_tags)
+            assert not ver[1].endswith(dirty_tag)
         assert ver[0] == exp_version
 
 
--- a/tests/test_install.py
+++ b/tests/test_install.py
@@ -183,26 +183,24 @@ class DemoApp:
 
 
 def check_version(version, exp_version, dirty=False):
-    dirty_tag = ".d" + strftime("%Y%m%d")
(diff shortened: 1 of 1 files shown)


FlorianWilhelm:
Thanks @abravalheri. Sorry to be so blunt, but this still feels a bit hacky as we're not really solving the main problem here, just alleviating the symptoms. My guess is that `setuptools-scm` relies on UTC where `time.strftime` uses the local time. So couldn't we just use instead:
```python
from datetime import datetime

datetime.utcnow().strftime("%Y%m%d")
```

abravalheri:
So that is what this problem is about!!

Thanks @FlorianWilhelm, for some reason this explanation completely slipped my mind 😅 

I would change the PR later today!

abravalheri:
Hi @FlorianWilhelm, I updated the implementation.

Doubt here: should we still be a bit loose in the comparison, just to cover the cases for example when the commit is generated just after midnight but the comparison happens slightly after 0:00?

abravalheri:
On a site note, now I am confused, in `setuptools-scm` code base there is one mention to UTC:
https://github.com/pypa/setuptools_scm/blob/088c652ec755f0982b3705272a1e6aa3e07c71ef/src/setuptools_scm/version.py#L116

but sometimes it uses the local date:
https://github.com/pypa/setuptools_scm/blob/088c652ec755f0982b3705272a1e6aa3e07c71ef/src/setuptools_scm/version.py#L336

tests/test_install.py
```diff
@@ -183,24 +183,20 @@ def tag(self, name, message):
 
 
 def check_version(version, exp_version, dirty=False):
-    dirty_tag = ".d" + strftime("%Y%m%d")
+    today = datetime.now(timezone.utc).strftime("%Y%m%d")
```
abravalheri, line 186: [Python docs](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow) recommend using explicit timezone instead of `utcnow`.

FlorianWilhelm:
> On a side note, now I am confused: in `setuptools-scm` they seem to use UTC once: https://github.com/pypa/setuptools_scm/blob/088c652ec755f0982b3705272a1e6aa3e07c71ef/src/setuptools_scm/version.py#L116
> 
> but other times they use the local date: https://github.com/pypa/setuptools_scm/blob/088c652ec755f0982b3705272a1e6aa3e07c71ef/src/setuptools_scm/version.py#L336

Maybe that's a bug in their code?

tests/test_install.py
FlorianWilhelm: Thanks, didn't know that.

FlorianWilhelm (approved):
Thanks, looks good and much cleaner.

abravalheri:
Thank you @FlorianWilhelm!

I asked about it in the discord, let see what happens...

abravalheri:
> Maybe that's a bug in their code?

Fixed upstream 🎉

PR #566, opened 2022-01-11
Fix pre-commit repo initialisation problem due to unsupported protocol

### LINKED ISSUE #565: Initializing pre-commit environment fails due to no longer supported unauthenticated protocol

pre-commit fails when initializing with:
```
[WARNING] The 'rev' field of repo 'https://github.com/psf/black' appears to be a mutable reference (moving tag / branch).  Mutable references are never updated after first install and are not supported.  See https://pre-commit.com/#using-the-latest-version-for-a-repository for more details.  Hint: `pre-commit autoupdate` often fixes this.
[INFO] Initializing environment for git://github.com/pre-commit/pre-commit-hooks.
An unexpected error has occurred: CalledProcessError: command: ('/usr/lib/git-core/git', 'fetch', 'origin', '--tags')
return code: 128
expected return code: 0
stdout: (none)
stderr:
    fatal: remote error:
      The unauthenticated git protocol on port 9418 is no longer supported.
    Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.

Check the log at /home/jovyan/.cache/pre-commit/pre-commit.log
```

This is already reported in an [issue of pyscaffoldext-dsproject](https://github.com/pyscaffold/pyscaffoldext-dsproject/issues/47).

FlorianWilhelm:
This should fix #565, thanks @daniel-smit-haw for reporting.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 exclude: '^(docs/conf.py|docs/gfx)'
 
 repos:
-- repo: git://github.com/pre-commit/pre-commit-hooks
+- repo: https://github.com/pre-commit/pre-commit-hooks
   rev: v4.1.0
   hooks:
   - id: trailing-whitespace
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,12 +14,10 @@ Changelog
 Current versions
 ================
 
-Version 4.1.2, 2022-01-22
+Version 4.1.3, 2022-01-XX
 -------------------------
 
-- Added verification to prevent users from creating projects nested inside
-  existing repositories, unless ``--force`` is explicitly given, :issue:`544`, :pr:`545`.
-- Fixed problem with Windows Shell when running git, :pr:`560`.
+- Fix pre-commit error: The unauthenticated git protocol on port 9418 is no longer supported, :issue:`565`
 
 .. note::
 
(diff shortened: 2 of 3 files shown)


abravalheri:
Thank you @FlorianWilhelm for spotting this!

Is there anyway we can tell Github to merge the PR once the tests pass?

FlorianWilhelm:
Thanks @abravalheri, I don't know if automatic merges are possible, surely with some kind of hook I guess or with Github Actions. I'll have a look and manually merge it.

Do you also think we should release directly 4.1.3 as this will otherwise let a lot of unittests fail in our extensions?

abravalheri:
> I'll have a look and manually merge it.

Thank you very much @FlorianWilhelm!

> Do you also think we should release directly 4.1.3 as this will otherwise let a lot of unittests fail in our extensions?

Definitely! Thank you very much for handling this and providing quick fixes!

abravalheri:
> I don't know if automatic merges are possible, surely with some kind of hook I guess or with Github Actions.

I found this one https://mergify.com. What do you think? Should we give it a try?

FlorianWilhelm:
Hm... to be honest, I like clicking merge and delete branch, one of my kinks 😅  . Also from a practical side. What if I realise that I want to add some more commits even if the current commit is already green? Also the number of parallel PRs is quite often really low, so an additional tool might add more complexity than the current situation. It seems https://mergify.com might help companies where dozens of teams work on a codebase with tons of PRs but in our situation... I am a bit sceptical to be honest.

abravalheri:
No problems! It is one of those tools that bring a very marginal benefit any way (I was just being lazy to get back to the PR once the CI finishes running...)

Issue #567, opened 2022-01-12
Increase awareness about workflow with `setuptools-scm` and `include_package_data=True`


abravalheri:
## Describe your use-case
_Originally described by @nickarini in pyscaffold/pyscaffoldext-custom-extension#49_

The way `setuptools-scm` works might contradict some users expectations: when they add a new file, they want to see it immediately included in the distribution.

This is a practical (and fair) limitation of the way `setuptools`/`setuptools-scm` works: you need to `git add` the file first (I think that is enough, but maybe you also have to do a commit - I quite don't remember now)

There is not much we can do about it, except to increase awareness (so users can adapt their workflow).

## Describe the solution you would like to see for your use-case
It would be nice to add a note about that in the docs.

Question: is it worthy to print some information directly on the terminal once putup finish running?

PR #568, opened 2022-01-12
More focus on tracking data files with git

FlorianWilhelm:
This addresses issue #567.


abravalheri (approved):
Thank you very much for handling this @FlorianWilhelm!

Nice touch converting the titles to appear in the navigational ToC!

I think at some point (not necessarily in this PR) it would be nice to add an entry to the FAQ.

PR #569, opened 2022-01-12
Add entry to FAQ about files not included in the distribution

### LINKED ISSUE #567: Increase awareness about workflow with `setuptools-scm` and `include_package_data=True`

abravalheri:
## Purpose
This is an incremental change on top of #568 so we can manage user's expectations about how files are included in the distribution.
Closes #567.

## Approach
- Add a FAQ entry that it is easily reachable by users before they open a new issue.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -166,6 +166,22 @@ What is the license of the generated project scaffold? Is there anything I need 
     is 0BSD-licensed and consequently you have no obligations at all and can do whatever you want except of suing us ;-)
 
 
+Why my file is not being included in the sdist/wheel distribution?
+    By default projects generated with PyScaffold rely on :pypi:`setuptools-scm` to populate the generated sdist or wheel, which in
+    turn uses ``git`` to list all the *non-transient project files*.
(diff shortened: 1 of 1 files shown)


abravalheri:
The rendered file can be seen at https://pyscaffold--569.org.readthedocs.build/en/569/faq.html

Is it OK to leave the question under the existing `Pyscaffold Usage` section, or should we add a new one?

FlorianWilhelm (approved):
Thanks @abravalheri! I think this will help a lot of people.

PR #571, opened 2022-01-13
Build sdists by default again

### LINKED ISSUE #570: Conda packages get no longer updated

There are no new packages on conda-forge for versions `4.1.2` and `4.1.3`. We should fix this.

The problem seems to be that we don't publish source distributions anymore to pypi. For instance, looking under
[version 4.1.1 files](https://pypi.org/project/PyScaffold/4.1.1/#files) there is a `tar.gz` but not under [version 4.1.2 files](https://pypi.org/project/PyScaffold/4.1.2/#files) and [version 4.1.2 files](https://pypi.org/project/PyScaffold/4.1.3/#files) anymore. I couldn't find a change in our build configuration from 4.1.1 that explains this. So no idea :-( 

Those files are needed in [meta.yml](https://github.com/conda-forge/pyscaffold-feedstock/blob/master/recipe/meta.yaml).

abravalheri:
## Purpose
Fix problem with conda packages.
Closes #570

## Approach
- Start building sdist packages, that will in turn be published to PyPI again.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,14 +14,10 @@ Changelog
 Current versions
 ================
 
-Version 4.1.3, 2022-01-11
+Version 4.1.4, 2022-01-13
 -------------------------
 
-- Fix pre-commit error: The unauthenticated git protocol on port 9418 is no longer supported, :issue:`565`
-
-.. note::
-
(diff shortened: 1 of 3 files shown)


FlorianWilhelm (approved):
Thanks, lgtm

PR #572, opened 2022-01-13
Remove advice about `sdists` from the docs

abravalheri:
## Purpose
Since now we found a good reason why using `sdist` is helpful (building conda packages), and it seems that the problems with `include_package_data` are solved in setuptools, I believe we no longer have to advise against them.

## Approach
- Simply remove the phrases in the docs advising against sdists
- Rely on `build` defaults

## Resources & Links

- [Experiments showing differences between sdists and wheels](https://github.com/abravalheri/experiment-setuptools-package-data#results-for-setuptools5920)

### DIFF
--- a/docs/features.rst
+++ b/docs/features.rst
@@ -25,9 +25,7 @@ compliant way, by just running::
     tox -e build
 
 Alternatively, if you are not a huge fan of isolated builds, or prefer running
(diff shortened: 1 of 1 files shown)


abravalheri:
Is there any other problem with setuptools options for sdists?
I am happy to investigate and submit a patch to setuptools if necessary.

PR #573, opened 2022-01-15
Test more reliably if extension is EntryPoint

FlorianWilhelm:
## Purpose
On the conda-forge [PR to release version 4.1.4](https://github.com/conda-forge/pyscaffold-feedstock/pull/44), we see a problem with the `test_iterate_entry_points` [unittest](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=441006&view=logs&j=656edd35-690f-5c53-9ba3-09c10d0bea97&t=e5c8ab1d-8ff9-5cae-b332-e15ae582ed2d).

It seems the problem is that for Python versions greater than 3.8 we are checking for `importlib.metadata.EntryPoint` while setuptools seems to use `importlib_metadata.EntryPoint` (note the underscore) all of a sudden. The test was slightly altered to check for the right string name of the class.  One could even go so far an just remove the `assert` line as basically we are checking external behavior.

### DIFF
--- a/tests/test_extensions.py
+++ b/tests/test_extensions.py
@@ -82,9 +82,6 @@ def test_iterate_entry_points():
     ext_iter = extensions.iterate_entry_points()
     assert hasattr(ext_iter, "__iter__")
     ext_list = list(ext_iter)
-    # isinstance(...) doesn't work reliably as we don't know if setuptools using
-    # importlib.metadata or importlib_metadata for Python >= 3.8
-    assert all([e.__class__.__name__ == "EntryPoint" for e in ext_list])
     name_list = [e.name for e in ext_list]
     for ext in ("cirrus", "pre_commit", "no_skeleton", "namespace", "venv"):
         assert ext in name_list
--- a/tests/test_extensions.py
+++ b/tests/test_extensions.py
@@ -82,7 +82,9 @@ def test_iterate_entry_points():
     ext_iter = extensions.iterate_entry_points()
     assert hasattr(ext_iter, "__iter__")
     ext_list = list(ext_iter)
-    assert all([isinstance(e, EntryPoint) for e in ext_list])
+    # isinstance(...) doesn't work reliably as we don't know if setuptools using
+    # importlib.metadata or importlib_metadata for Python >= 3.8
+    assert all([e.__class__.__name__ == "EntryPoint" for e in ext_list])
     name_list = [e.name for e in ext_list]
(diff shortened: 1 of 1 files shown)


abravalheri:
Thank you very much @FlorianWilhelm.

Maybe we should embrace duck-typing and assert `hasattr(e, 'load')` instead? (Which is the method we use and would make sense to have the test suite protecting us from regressions...). Or just dropping the check as you suggested.

---

A comment: I don't understand this error, both library module and test use the same condition to import `EntryPoint`-related functions and classes, so there should be no clash, right?

tests/test_extensions.py
```diff
@@ -82,7 +82,9 @@ def test_iterate_entry_points():
     ext_iter = extensions.iterate_entry_points()
     assert hasattr(ext_iter, "__iter__")
     ext_list = list(ext_iter)
-    assert all([isinstance(e, EntryPoint) for e in ext_list])
+    # isinstance(...) doesn't work reliably as we don't know if setuptools using
```
abravalheri, line 85: We should no longer be relying on setuptools/pkg_resources for the entry-points, right?

And both library module and test do the same conditional import...

That is very tricky!

abravalheri:
Maybe the problem is that the recent version of `importlib-metadata` changed the signature?

```python
def entry_points(**params) -> Union[EntryPoints, SelectableGroups]: ...
```

So it is no longer guaranteed to have an `EntryPoint`?

https://github.com/python/importlib_metadata/blob/9491ef9a2733a7891c665c0c7969804a3567dbe3/importlib_metadata/__init__.py#L981-L1003

FlorianWilhelm:
@abravalheri, I could first not reproduce this locally by then using `environment.yaml` to setup a fresh new conda environment the same errors as on azure pipelines appeared. I did some debugging and found out that even though I was using Python 3.10 the actual `EntryPoint` was from `importlib_metadata` as I verified using `type(...)`. Since this all worked fine in my old conda environment I thought that it must somehow be related to the setuptools toolchain (but this is only a guess). 

So how do we proceed? I think just dropping this one assert is fine, right? Below that, we check for some attributes of `EntryPoint` that we also use in the code and if somehow this was no `EntryPoint` we would also fail (and see that something changed).

abravalheri:
> I think just dropping this one assert is fine, right?

Yeah let's do this...

Thank you very much for investigating this problem @FlorianWilhelm. Please feel free to remove the test and merge (or merge as it is right now)

PR #574, opened 2022-01-17
Mention `python -m setuptools_scm` as a form of debugging

abravalheri:
## Purpose
Previously our docs mentioned the command `python setup.py version` as a form of debugging the version string. For a while this command was considered deprecated (thanks to PEP 517/518 and the changes in the ecosystem), so we decided to omit it.

Now that the [new version of `setuptools-scm` is available](https://github.com/pypa/setuptools_scm/releases/tag/v6.4.0), we have a replacement 🎉 

## Approach
- Mention `python -m setuptools_scm` as a form of obtaining the version string
- I also added an entry to FAQ about that...


FlorianWilhelm:
Nice! One step closer to getting rid of `setup.py` :-)

PR #576, opened 2022-01-18
Add pyenv's .python-version into .gitignore

prassanna-ravishankar:
## Purpose
Some people use pyenv local, which introduces in a .python-version in the cwd. This is susceptible to be tracked by git

## Approach
Adds .python-version to gitignore

## Resources & Links

Debatable whether this is the best approach (https://stackoverflow.com/questions/54315206/should-we-gitignore-the-python-version-file#comment95589915_54315206) . This has arguably caused more problems than it solves as different contributors have different virtual environment management mechanisms


FlorianWilhelm:
Thanks, @prassanna-ravishankar!

Issue #562, opened 2022-01-04
Add some more debugging information about called functions for `--very-verbose`-mode.


FlorianWilhelm:
In issue #556 we have seen that `--very-verbose`-mode can be hard to interpret sometimes. For instance there are several
```
         run      git.exe rev-parse --git-dir
    last command failed with "fatal: not a git repository (or any of the parent directories): .git"
```
in the output. Only by looking in the code one finds out that this is done in the `is_git_repo` function. So we could add debug statements to functions like `is_git_repo` but the question is where? Two alternatives:

1. Within `is_git_repo` like "Checking if {cwd} is git repository" or
2. rather outside where we could even state our intention like "Making sure that parent directory {dir} is no git repo".

Alternative 1. is more straightforward, less code and even less error-prone as you cannot forget to add a debug statement when another `is_git_repo` call is introduced. Alternative 2. provides more information about the intention, which is only available outside, and could thus be more informative when looking for errors.

Not sure what the best practice is but a few more debug statements could help. The goal should be that the output of `putup --very-verbose my_test_proj` is easy to interpret at least for us developers.

abravalheri:
Oops, sorry for creating a release before we had the chance to consider this issue 😅 

I did a quick search and apparently `is_git_repo` is only called in 2 places 

`actions.verify_project_dir`
https://github.com/pyscaffold/pyscaffold/blob/0dbaf83f35bbd9546dc5174f01f89b33338bebed/src/pyscaffold/actions.py#L303

and  `actions.init_git`
https://github.com/pyscaffold/pyscaffold/blob/0dbaf83f35bbd9546dc5174f01f89b33338bebed/src/pyscaffold/actions.py#L320

So going for option 2 is still viable... Would it make sense to pass a `log` argument with a custom message to the `is_git_repo` function? If none is given we can fallback to the generic/non-informative message...

FlorianWilhelm:
No problem, I already thought that this could go in the next release.

If you also like alternative 2 I would take a look and prep a PR. It will not only be `is_git_repo` but also other functions like it. Basically all functions that execute shell commands.

abravalheri:
Thank you @FlorianWilhelm

FlorianWilhelm:
Resolved in the PR above.

PR #578, opened 2022-01-24
Added more debugging information

FlorianWilhelm:
## Purpose

This adds more debugging information especially when calling shell commands as described in issue #562.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,7 @@ Version 4.1.5, 2022-XX-XX
 -------------------------
 
 - Added ``.python-version`` created by ``pyenv`` to ``.gitignore``
+- Added more ``--very-verbose`` logging to make clearer what's going on, :issue:`562`
 
 
 Version 3.3, 2020-12-24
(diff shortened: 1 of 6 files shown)


abravalheri:
Thank you very much @FlorianWilhelm, looks good to me:

image

I created a PR that I believe would fix the problems you see here: #579 🤞

abravalheri (approved):

FlorianWilhelm:
@abravalheri, I think using `logger.report` looks a bit more consistent for checks and verifications:
image

FlorianWilhelm:
> Thank you very much @FlorianWilhelm , it does look great!
> 
> I suppose the next step would be also using `record` for that extra spourious line right (the one starting with `last command`)?
> 
> After we merge this PR, I can try to give it a try.

Hm... right. I wasn't sure which `action` to pick as we are only returning the output of the command. Do you have a suggestion?

abravalheri:
"output" seems good to me, or just "failed"...

FlorianWilhelm:
@abravalheri, I opted now for `info`. What do you think?
image

abravalheri:
Thank you very much @FlorianWilhelm . It looks very good!

Please feel free to merge ;)

PR #579, opened 2022-01-24
Fix `find_report` test helper

abravalheri:
@FlorianWilhelm this should fix the problems we are seeing on #578.

### DIFF
--- a/tests/log_helpers.py
+++ b/tests/log_helpers.py
@@ -5,7 +5,8 @@ import re
 def find_report(log, activity, subject):
     """Check if an activity was logged."""
     for record in log.records:
-        if record.activity == activity and str(subject) in str(record.subject):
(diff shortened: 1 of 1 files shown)


FlorianWilhelm:
Hi @abravalheri, cool thanks! I was wondering if I should rather use `logger.report` instead of `logger.debug` anyway, what do you think? Maybe the debug output would look cleaner?

abravalheri:
Currently the output looks like:

image

If I am not wrong, what `logger.report` does is to align the "activity" right and to add hierarchical indentation (e.g. in the figure above `git --version` is called from inside `pyscaffold.actions:get_default_options` so it gets indented).

I don't know if the current behaviour is considered clean, but if you like, please go ahead 😝, otherwise we can also discuss how to improve it.

PR #583, opened 2022-02-01
Update versions of containers in CI config

abravalheri:
## Purpose
I noticed in the last PRs that the `coverage` task is failing. This is likely to be caused by some packages used in `pre-commit` dropping support for 3.6, so I decided to update our CI configs.

## Approach
- Update `coverage` task to use 3.8 (this is the default version in Ubuntu 20.04LTS, so I believe a lot of people would be using it).
- Remove 3.6 support in the Cirrus CI template (this version reached EoL the end of last year)
- Add 3.11 with `allow_failures` just for PyScaffold (not the template), so we start seeing if something will fail.
- Keep 3.6 on PyScaffold's own config, but add `allow_failures` (Ubuntu 18.04LTS still uses it, so even if we decide to not fix the bugs, we can know about them)

- In our Cirrus config file we have the following comment:
```yaml
    - name: test (OS X)
      macos_instance: {image: "big-sur-xcode"}
      env:
        PYTHON_VERSION: 3.7
        # ^  update when the default version of python in homebrew changes
```

@FlorianWilhelm do you know if 3.7 is still the version you get when you `brew install python`?
I could not find this information online... https://formulae.brew.sh/formula/python@3.7 does not say anything about this being the default version and if I remove the `@3.7` from the URL I get a 404 error...

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -64,9 +64,9 @@ env:
 # ---- Task definitions ----
 
 typecheck_task:
-  name: typecheck (Linux - 3.9)
+  name: typecheck (Linux - 3.10)
   clone_script: *clone
-  container: {image: "python:3.9-buster"}  # most recent => better type support
+  container: {image: "python:3.10-bullseye"}  # most recent => better type support
   pip_cache: *pip-cache
   mypy_cache:
     folder: .mypy_cache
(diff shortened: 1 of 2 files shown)


FlorianWilhelm:
@abravalheri. I just tried `brew install python` and it even wants to update to Python 3.9.10.

abravalheri:
Thank you very much @FlorianWilhelm, I submitted another commit with the update. Let's see if the CI jobs pass.

PR #584, opened 2022-02-04
Replace `appdirs` with `platformdirs`

### LINKED ISSUE #581: Note to our future selves: it seems that the community is moving past `appdirs` and that a more maintained fork has appeard: `platformdirs`.
It seems that the community is moving past `appdirs` and that a more maintained fork has appeared: `platformdirs`.

Ref:
- https://github.com/ActiveState/appdirs/issues/79
- https://github.com/ActiveState/appdirs/issues/170

So for the next version we might consider replacing the dependency (which should be API compatible).

_Originally posted by @abravalheri in https://github.com/pyscaffold/pyscaffold/issues/580#issuecomment-1025542512_

abravalheri:
## Purpose
As discussed in #581, the community is moving past `appdirs` and towards `platformdirs`.
The idea of this PR is to replace this dependency with the more maintained one.

Closes #581.

## Approach
In the current release `platformdirs` is designed to have a backwards compatible API, so the approach I took in this PR is to blindly replace everything.

I also replaced the mentions in examples and documentations to promote the more maintained alternative.

### DIFF
--- a/docs/features.rst
+++ b/docs/features.rst
@@ -124,7 +124,7 @@ website.
    considered good practice. Package files should be read-only. If you need
    configuration files, or files that should be written at runtime, please
    consider doing so inside standard locations in the user's home folder
-   (`appdirs`_ is a good library for that).
+   (:pypi:`platformdirs` is a good library for that).
    If needed you can even create them at the first usage from a read-only
    template, which in turn can be a package file.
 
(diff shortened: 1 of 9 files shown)


abravalheri:
@FlorianWilhelm do you think this could be framed as a minor update? I don't think it changes the APIs...

FlorianWilhelm:
Thanks @abravalheri! Yes, I guess it's rather minor.

PR #587, opened 2022-02-08
Simplify coverage config template

abravalheri:
## Purpose

This is a proposal to simplify the `.coveragerc` template, that takes advantage of advancements in recent versions of coverage:

- Good default values for `exclude_lines`.

I also would like to use `source_pkgs` instead of `source` since it is a more straight-forward option. However `source_pkgs` seems to be problematic with implicit namespaces...

---

There is a duplication between `.coveragerc` and `setup.cfg`:

a. `.coveragerc` contains `souce = ${package}`
b. `setup.cfg` contains `--cov ${qual_pkg}`.

Ideally it would be nice to avoid double bookkeeping.

## Approach

In this commit I am proposing centralising all the settings in`.coveragerc`, but it would be equally nice to make `.coveragerc` package agnostic and leave the varying parameter to be specified in `setup.cfg`... I have no strong preference here.

I removed all `exclude_lines` configuration because they seem to be covered by the defaults.

## Resources & Links

I did 2 experiments to test this change:

- pyscaffold/pyscaffoldext-cookiecutter#17
- pyscaffold/pyscaffoldext-markdown#43

### DIFF
--- a/src/pyscaffold/templates/coveragerc.template
+++ b/src/pyscaffold/templates/coveragerc.template
@@ -2,7 +2,6 @@
 # See https://coverage.readthedocs.io/en/stable/config.html
 [run]
 branch = True
-source = ${qual_pkg}
 # omit = bad_file.py
 
 [paths]
--- a/src/pyscaffold/templates/coveragerc.template
+++ b/src/pyscaffold/templates/coveragerc.template
@@ -2,6 +2,7 @@
 # See https://coverage.readthedocs.io/en/stable/config.html
 [run]
 branch = True
+source = ${qual_pkg}
 # omit = bad_file.py
 
 [paths]
--- a/src/pyscaffold/templates/coveragerc.template
+++ b/src/pyscaffold/templates/coveragerc.template
@@ -1,28 +1,12 @@
 # .coveragerc to control coverage.py
+# See https://coverage.readthedocs.io/en/stable/config.html
 [run]
 branch = True
-source = ${package}
+source = ${qual_pkg}
 # omit = bad_file.py
 
 [paths]
+# Paths that should be considered equivalent
 source =
     src/
     */site-packages/
-
-[report]
-# Regexes for lines to exclude from consideration
-exclude_lines =
-    # Have to re-enable the standard pragma
-    pragma: no cover
-
-    # Don't complain about missing debug-only code:
-    def __repr__
-    if self\.debug
-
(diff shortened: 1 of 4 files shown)


abravalheri:
The main drawback here is that `raise NotImplementedError` is no longer ignored.

There is [this package](https://pypi.org/project/covdefaults/) that could also be used...

@FlorianWilhelm do you have any thoughts?

src/pyscaffold/templates/coveragerc.template
```diff
@@ -1,28 +1,12 @@
 # .coveragerc to control coverage.py
+# See https://coverage.readthedocs.io/en/stable/config.html
 [run]
 branch = True
-source = ${package}
+source = ${qual_pkg}
```
FlorianWilhelm, line 5: This was a bug for namespace packages, right?

FlorianWilhelm (approved):
Thanks, it surely simplifies things, and if everything except `raise NotImplementedError` is now a default in coverage it makes sense. 
I have mixed feelings regarding the coverage plugin for sane defaults as this just hides things, right? On the other hand one would leave sane settings to external that take care of it. Hm.... what do you think? We could first merge this PR and evaluate the plugin?

Regarding your pytest-virtualenv comment. We still need it for some unit tests, right? It used to be quite a pain or can we get rid of it now easily? (Without reimplementing what it does)

abravalheri:
One good aspect of PyScaffold how it is implemented right now its that it has this "educational" aspect to its templates. The user starts using the templates without having to know them or understand the tools, but as soon as they need to change they can learn from the examples and adapt to their needs.

On the other hand, one of the motivations for me to reduce the size of the coverage configuration is that the plans for v5.x include merging this config into `pyproject.toml`, and having a massive sized configuration file is never good...

Adding the plugin would be a trade-off between a more "manegeable" `pyproject.toml` and an "educational" approach for `coverage`.

I am leaning towards the plugin because it is quite comprehensive, and I think the amount of times people will have to fine tune it is small.

FlorianWilhelm:
@abravalheri, okay, then let's try the plugin :-) Thanks for this PR!

abravalheri:
I just tried it in the [ConfigUpdater repository](https://github.com/pyscaffold/configupdater/pull/86#issuecomment-1046086448) and I got a bit disappointed because it forces setting up a threshold for the coverage :(

I think I will abandon this PR and submit the fix for the `${qual_pkg}` separately.

Issue #605, opened 2022-02-13
PreCommit should fix mixed-line-endings automatically / Add .gitattribute


julie777:
### Discussed in https://github.com/pyscaffold/pyscaffold/discussions/337

<div type='discussions-op-text'>

<sup>Originally posted by **CarliJoy** December 14, 2020</sup>
## Description of your problem

Copy stuff from somewhere to PyCharm from a source with different line ending.

This will prevent me from committing at all.
Users with not so much experience (not everybody who writes python packages is a computer geek) will have a hard time getting the commit to work.
It would be nice if the setting is more forgiving, so there is an error but an auto fix.

**Please provide any additional information below.**
Suggestion: Change the `.pre-commit-config.yaml` to
```
  - id: mixed-line-ending
    args: ['--fix=auto']
```

I don't know if that catches all edge cases especially working with autoclrf.
Maybe it would be nice to introduce a .gitattribute instead?
See also: https://stackoverflow.com/questions/2825428/why-should-i-use-core-autocrlf-true-in-git

It probably depends also a lot on the user settings but maybe we still could safely add some files (like *.sh) to a gitattribute file?

PS: Github offers Discussions now. This issue would fit right into them.</div>

julie777:
I would like to implement the solution I proposed in  the discussion

Implementing The Requirement

A good way to ensure the requirement is met, and to make sure it doesn't depend on precommit or local developer settings it to add the following two files to the project.

    With this config-based approach, you can ensure that your line endings remain consistent in your codebase regardless of what operating systems or local Git settings your developers use since this file takes priority.

.gitattributes

* text=auto

    Many code editors such as vscode will use this file if present to keep workspace file line endings correct.

.editor.config

root = true

[*]
end_of_line = lf

(If we decide to do this, I will gladly add the new files and make sure all existing files in the repo are normalized.)

julie777:
Currently the only file in the repo with non LF endings is docs/gfx/logo.svg which has CRLF endings.
Given that all the other .svg files have LF line endings I don't imagine that changing logo.svg to LF endings will cause any problems.

abravalheri:
Hi @julie777, please find bellow my attempt to summary of the presented arguments in favour of the change is (from what I could collect from the comments and links):

1. People have trouble with git's `autocrlf=True` configuration as it leaves the local repository in a different state than the remote. Using `.gitattributes` seem to offer a different (better) experience than `autocrlf=True`
[...]

PR #606, opened 2022-02-13
issue #605 added support to force line endings to LF for text files

julie777:
also fixed the one file with the wrong line endings

## Purpose

## Approach

## Resources & Links


julie777:
added another commit for a missed file change. These can be squashed when merged.

julie777:
This is my first time going through this specific process. I'm not sure how to squash my commits into a single commit or if that happens during the merge of the pull request.

abravalheri:
Hi @julie777, thank you very much for looking into this. And congratulations for your first PR!

(Sorry, I still haven't had the time to reply to your other messages/comments/issues, but I promise I will get to that soon!)

Can you elaborate a bit more why the current solution (`args: ['--fix=lf']` in `.pre-commit-config.yml`) does not work?

(I am disregarding the situations when PyScaffold devs forget to run `pre-commit` install. In those situations isn't it a good thing to have some discrepancies so it serves as reminder to do the installation?)

julie777:
> Can you elaborate a bit more why the current solution (args: ['--fix=lf'] in .pre-commit-config.yml) does not work? (Or at least in which situations it would not work).

Mostly because it is built in to git and doesn't require and hooks that might not get run or not run until CI. 
I think the doc I linked might have a more detailed explanation.

FlorianWilhelm:
Thanks for this PR @julie777. This PR introduces `.editorconfig` and I wonder if we should also add more things like
```
[*.py]
indent_style = space
indent_size = 4
```
and if maybe later it should also be part of the PyScaffold template?

The fix for the line endings in the SVG graphic is fine. Regarding the `.gitattributes` I am also not 100% as I am actually not using them so often. I found this [Stackoverflow question](https://stackoverflow.com/questions/21472971/what-is-the-purpose-of-text-auto-in-gitattributes-file) and at least here the recommendation is not to use it as Git might make a mistake and convert non-text files. Then there is [this question](https://stackoverflow.com/questions/33872793/git-text-auto-in-the-gitattributes-file-and-line-endings?noredirect=1&lq=1) from a Windows users confused about the effects of Git's automatic conversion from the repo to the working directory. I am just not sure if adding this really helps. Shouldn't it be enough to have it in `.editorconfig` and `pre-commit`?

PR #619, opened 2022-02-26
Add builtin extension for GitHub actions

### LINKED ISSUE #589: Add Github Actions CI/CD
## Describe your use-case
Github Actions seems to become more and more adapted for CI/CD, it thus could be nice to have besides `--gitlab` and `--cirrus` also a `--github` flag.

abravalheri:
New extension to generate a configuration file for a GitHub Actions workflow.

Although the template I am introducing here is not the most compact and simple to understand in the world, I think it is very complete!

You can see it in action in https://github.com/pyscaffold/ci-tester/pull/9, [ci logs](https://github.com/pyscaffold/ci-tester/runs/5350808516?check_suite_focus=true) and https://test.pypi.org/project/ci-tester/0.0.0.0.0.1a2/

Please feel free to add any suggestions.

Closes #589.

### DIFF
--- a/src/pyscaffold/templates/github_ci_workflow.template
+++ b/src/pyscaffold/templates/github_ci_workflow.template
@@ -34,7 +34,7 @@ jobs:
       - uses: actions/setup-python@v2
         with: {python-version: "3.10"}
       - name: Lint code
-        run: pipx run tox -e lint
+        run: pipx run pre-commit run --all-files --show-diff-on-failure
       - name: Build package distribution files
         run: pipx run tox -e clean,build
       - name: Record the path of wheel distribution
--- a/src/pyscaffold/templates/tox_ini.template
+++ b/src/pyscaffold/templates/tox_ini.template
@@ -20,15 +20,17 @@ commands =
     pytest {posargs}
 
 
-[testenv:lint]
-description = Perform static analysis and style checks
-skip_install = True
-deps = pre-commit
-passenv =
-    HOMEPATH
-    PROGRAMDATA
-commands =
-    pre-commit run --all-files {posargs:--show-diff-on-failure}
+# # To run `tox -e lint` you need to make sure you have a
+# # `.pre-commit-config.yaml` file. See https://pre-commit.com
+# [testenv:lint]
+# description = Perform static analysis and style checks
+# skip_install = True
+# deps = pre-commit
+# passenv =
+#     HOMEPATH
+#     PROGRAMDATA
+# commands =
+#     pre-commit run --all-files {posargs:--show-diff-on-failure}
 
 
 [testenv:{build,clean}]


src/pyscaffold/templates/tox_ini.template
```diff
@@ -20,6 +20,17 @@ commands =
     pytest {posargs}
 
 
+[testenv:lint]
```
abravalheri, line 23: Maybe this testenv should not be added by default to every user, considering that `pre-commit` will not be installed by everyone (so the pre-commit hooks file will be absent)...

src/pyscaffold/templates/github_ci_workflow.template
```diff
@@ -0,0 +1,114 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      wheel-distribution: ${{ steps.wheel-distribution.outputs.path }}
+    steps:
+      - uses: actions/checkout@v2
+        with: {fetch-depth: 0}  # deep clone for setuptools-scm
+      - uses: actions/setup-python@v2
+        with: {python-version: "3.10"}
+      - name: Lint code
+        run: pipx run tox -e lint
```
abravalheri, line 37: Here we can just `pipx run pre-commit --all-files` instead (considering that the `testenv:lint` environment is not available).

CHANGELOG.rst
```diff
@@ -10,6 +10,14 @@ Changelog
     Version 4.X.X, 2022-XX-XX
     -------------------------
 
+Development version
+===================
+
+Version 4.2, 2022-XX-XX
+-----------------------
+
+- Added ``--github-actions`` extension, :pr:`619`
```
abravalheri, line 19: We could also name the flag `--gh-actions` instead... (I am bad at naming)

FlorianWilhelm: I think it's fine, thanks!

src/pyscaffold/templates/github_ci_workflow.template
FlorianWilhelm: Ok, then it's also clearer what happens and one doesn't need to look up what `lint` is mapped to. On the other hand, I got used to calling `tox -e build`, and using `tox` for all kinds of runner tasks is also fine. So both is fine for me. 
Why is the `testenv:lint` not available?

src/pyscaffold/templates/tox_ini.template
FlorianWilhelm: Hm... and if we let `github-actions` require `pre-commit`? Now I understand the comment before. Since `pre-commit` becomes so important, we could also in a 5.0 release switch to making it as default and providing a `no-pre-commit` flag instead. 

For the time being, does it hurt to have a `testenv:lint` that is never called if the user doesn't run it? In the `github_ci_workflow.template` one could add or not add the `lint`-step depending if `--precommit` was chosen for the time being. Or just adding `--precommit` in case `--github-actions` was chosen. I would argue that one always wants some linting checks when working with CI/CD.

src/pyscaffold/templates/tox_ini.template
```diff
@@ -60,10 +72,11 @@ description =
...
 passenv =
+    # See: https://twine.readthedocs.io/en/latest/
     TWINE_USERNAME
     TWINE_PASSWORD
     TWINE_REPOSITORY
 deps = twine
 commands =
     python -m twine check dist/*
-    python -m twine upload {posargs:--repository testpypi} dist/*
+    python -m twine upload {posargs:--repository {env:TWINE_REPOSITORY:testpypi}} dist/*
```
FlorianWilhelm, line 82: Cool, that gives us some more flexibility!

FlorianWilhelm (approved):
Thanks @abravalheri! Looks good to me. See my comments regarding the linting.

src/pyscaffold/templates/tox_ini.template
abravalheri: I agree with adding `pre-commit` by default in v5, however we have to discuss how *exactly* we are doing that[^1]

[^1]: Before v4, running `putup --pre-commit` had only the side effect of producing some extra file. Therefore, there was no problem in having it as a default (if the user didn't want to run `pre-commit --install` no changes in workflow would happen). Now, however, `putup --pre-commit` will try to automatically run `pre-commit install`, so leaving it as a default, without changing anything else will change the dev's workflow.

> does it hurt to have a `testenv:lint` that is never called if the user doesn't run it?

Currently there will be no problem for users running `--github-actions`, because this extension will also automatically activate `--pre-commit` even if `--pre-commit` is not passed. The situation without `--github-actions` is the one that changes.

What might happen if the default `tox.ini` template have a `testenv:lint` is:
- an user that runs `putup` without `--pre-commit`
- the user runs `tox -av` to find out which tasks are available
- the user sees `lint` there
- the user runs `tox -e lint`
- tox fails with the following error message:
  ```
  An error has occurred: InvalidConfigError:
  =====> .pre-commit-config.yaml is not a file
  Check the log at ~/.cache/pre-commit/pre-commit.log
  ```

abravalheri: In https://github.com/pyscaffold/pyscaffold/pull/493#issuecomment-959797846, https://github.com/pyscaffold/pyscaffold/pull/493#issuecomment-961381764 and https://github.com/pyscaffold/pyscaffold/pull/493#issuecomment-961385956 I was playing around with the idea of having an API were we can comment/uncomment lines of the files in the tree structure[^1]...

[^1]: While I agree that enabling general modifications to files is very risky and error-prone, I think the capability of commenting/uncommenting and adding/removing entire blocks of text is less problematic...

Using the same line of thought, one solution here would be having the `testenv:lint` commented out by default and then the `--pre-commit` extension could uncomment it (or simply adding it in the case it is not there in the form of a commented block).

Other viable solutions is to make the default `tox.ini` template "aware" of pre-commit. So when the `--pre-commit` extension is activated, the testenv is added to the tox template. (The downside here is increasing coupling)

abravalheri:
Thank you very much for the review and the nice discussion @FlorianWilhelm!

src/pyscaffold/templates/tox_ini.template
FlorianWilhelm: I think your "commenting and uncommenting"-solution is the best in this case. We should find a nice and robust way to do this.

abravalheri:
In https://github.com/pyscaffold/pyscaffold/pull/619/commits/112e9364ff75a165d1083b8292d661515ef5a13c I updated the PR to not introduce `tox -e lint` directly into the tox template, otherwise that would be problematic for users.

PR #622, opened 2022-03-08
Revamp `--gitlab` extension

abravalheri:
The previous version of the template for the GitLab CI was added to PyScaffold a while ago and never revised.

Although it seems that it was working (surprisingly), it also didn't offer the same features as CirrusCI or the proposed GitHub Actions.

This change is a revamp on that template that tries to offer more or less the same level of functionality our other CI configs offer.

It was tested against a replica of the `ci-tester` repository on GitLab:

- https://gitlab.com/pyscaffold/ci-tester

This is the pipeline output:

- https://gitlab.com/pyscaffold/ci-tester/-/commit/e8e3a777b972bc839a9b32d7c0c0678184ef2756

### DIFF
--- a/src/pyscaffold/extensions/gitlab_ci.py
+++ b/src/pyscaffold/extensions/gitlab_ci.py
@@ -1,18 +1,28 @@
 """
 Extension that generates configuration and script files for GitLab CI.
 """
-
+from argparse import ArgumentParser
 from typing import List
 
 from .. import structure
 from ..actions import Action, ActionParams, ScaffoldOpts, Structure
 from ..operations import no_overwrite
 from ..templates import get_template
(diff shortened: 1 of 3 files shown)


abravalheri:
Maybe in the future we want to change the flag from `--gitlab` to `--gitlab-ci`...
I fear that having it named `--gitlab` might influence people that are hosting the code on GitLab to use that flag, even if they would prefer having other CI provider (e.g. Cirrus)

FlorianWilhelm (approved):
Thanks, @abravalheri! LGTM :-)

abravalheri:
Thank you very much for the review @FlorianWilhelm!

PR #625, opened 2022-03-11
Update and simplify Cirrus CI config

### LINKED ISSUE #623: Improvements for Cirrus CI
After having worked on both the GitHub Actions and GitLab CI plugins, I think we also need to improve Cirrus CI 😅

It is a bit behind... The following would be "nice to have" improvements:
- [ ] Upload to coveralls for all tasks using different tags (this relies on Coveralls' "parallel" builds, there is an example on how to do that on the GitlabCI PR https://github.com/pyscaffold/pyscaffold/pull/622)
- [ ] Use the "pipeline" approach: build the package first, then use the same package to test across the different OS/Python versions
- [ ] Add support for publishing releases.
- [ ] It would also be nice to simplify that file... Currently it is too big and too complicated to understand (I guess the culprit is Windows)

abravalheri:
## Purpose

The config for Cirrus CI have always been big and complicated.
While being big is not a problem (as long as it is worthy in terms of features), it would be nice to avoid complicated configs.

I think the main culprit for the complicated config is Windows, so lately I have been investigating ways of simplifying it.

Moreover with the latest improvements in PyScaffold in terms of `--gitlab` and `--github-actions`, the configuration for `--cirrus` fell a bit behind. For example, the current configuration does not support:

- aggregating coverage metrics for parallel runs
- testing the same distribution that is produced with `tox -e build`
- publishing packages to PyPI

## Approach

This change bring the features mentioned above to `.cirrus.yml`. Moreover the following were also implemented:

- Added support for `freebsd` tests
- Test on `mambaforge` instead of Anaconda
- Simplified Windows config:
  - Changed default shell to `bash`
  - Changed installation method (Python docker + nuget => to Cirrus-provided docker + choco)

- Should we remove [`.github/workflows/publish-package.yml`](https://github.com/pyscaffold/pyscaffold/blob/master/.github/workflows/publish-package.yml) now that Cirrus can also upload packages?
  The advantage of using Cirrus is that we can upload exactly the same package that we run the test suite for.

----
Closes #623

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -40,8 +40,9 @@ env:
   depends_on: [build]
   <<: *task-template
   dist_cache: {folder: dist, fingerprint_script: echo $CIRRUS_BUILD_ID}  # download
-  test_script:
-    - tox --installpkg dist/*.whl -e all -- -n auto -rfEx --durations 10 --color yes
+  test_script: >
+    tox --installpkg dist/*.whl -e all --
+    -n auto --randomly-seed=42 -rfEx --durations 10 --color yes
   submit_coverage_script:
     - pipx run coverage xml -o coverage.xml
     - pipx run coveralls --submit coverage.xml
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -42,7 +42,7 @@ env:
   dist_cache: {folder: dist, fingerprint_script: echo $CIRRUS_BUILD_ID}  # download
   test_script: >
     tox --installpkg dist/*.whl -e all --
-    -n auto --randomly-seed=42 -rfEx --durations 10 --color yes
+    -n 5 --randomly-seed=42 -rfEx --durations 10 --color yes
   submit_coverage_script:
     - pipx run coverage xml -o coverage.xml
     - pipx run coveralls --submit coverage.xml
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -66,7 +66,7 @@ check_task:
   <<: *task-template
   install_script: pip install tox
   check_script:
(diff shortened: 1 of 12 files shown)


.cirrus.yml
```diff
@@ -98,130 +98,82 @@ linux_mac_task:
...
-    - python -m pip install --upgrade pip setuptools tox
-  prepare_script: *prepare
-  linkcheck_script:
-    - python -m tox -e linkcheck -- -q
+  dist_cache: {folder: dist, fingerprint_script: echo $CIRRUS_BUILD_ID}  # download
+  <<: *task-template
+  install_script: pip install tox
+  linkcheck_script: tox --installpkg dist/*.whl -e linkcheck -- -q
+
+# publish_task:
```
abravalheri, line 164: Here we can decide to uncomment this task, but then we should also remove the [`.github/workflows/publish-package.yml`](https://github.com/pyscaffold/pyscaffold/blob/master/.github/workflows/publish-package.yml) file otherwise we will end up trying to publish the package twice (which will result in error).

abravalheri:
@FlorianWilhelm do you have any thoughts about this change?
In terms of total number of lines reduced, it is not much (less than 40), but at the same time it is adding some functionality (e.g. FreeBSD tests, parallel coverage aggregation)...

.cirrus.yml
FlorianWilhelm: Yes, should we add a comment about this remark? So that we don't run into this problem in the future?

FlorianWilhelm:
Thanks @abravalheri! Looks more structured and cleaner.

PR #630, opened 2022-03-16
before_script needs to be an array

baurmatt:
## Purpose
Fix Gitlab CI pipeline

```
jobs:mamba:before_script config should be an array containing strings and arrays of strings
```

## Approach
Converting the YAML object type to an array

## Resources & Links

### DIFF
--- a/src/pyscaffold/templates/gitlab_ci.template
+++ b/src/pyscaffold/templates/gitlab_ci.template
@@ -54,12 +54,14 @@ default:
 check:
   stage: prepare
   image: "python:3.10"
-  script: pipx run pre-commit run --all-files --show-diff-on-failure
+  script:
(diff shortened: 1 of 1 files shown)


baurmatt:
@abravalheri This was a CI job running on a self hosted Gitlab Enterprise 14.7.4. Interesting that it works on gitlab.com 🤔

abravalheri:
Might be a version thing...

If that is the case, maybe the best is to use arrays for all the `scripts`, `before_scripts`, `after_scripts`, etc right? What do you think?

Would you be interesting in changing the other places as well?

abravalheri:
Amazing!

Thank you very much for all the help @baurmatt. I will merge this change, run some tests in our `ci-tester` repository on GitLab (just as a sanity check) and then release a patch version shortly.

PR #632, opened 2022-03-28
Update migration.rst

eacheson:
I think the author was trying to say "maybe", but "eventually" in English means "after some time" / "in the end"

## Purpose

## Approach

## Resources & Links


abravalheri:
Thank you very much @eacheson for spotting this and providing a fix.

PR #640, opened 2022-05-26
Change the pattern for versions in the github actions template

abravalheri:
Improve the pattern used to match version tags in the GitHub CI template.


sgbaird:
Dropped the ball on this from getting sick, and just getting back to it. Thanks for this!

PR #641, opened 2022-05-26
Change note in skeleton to emphasize it can be removed/renamed

abravalheri:
## Purpose
Improve note on `skeleton.py` to clarify the document can be safely removed/renamed.

## Approach

See discussion #638.


sgbaird:
Thanks!

Issue #649, opened 2022-06-17
Tests don't run after migrating to pyscaffold


ofosos:
## Description of your problem

After migrating to pyscaffold, I get some errors trying to run tox.
 
**Please provide a minimal, self-contained, and reproducible example.**

Checkout https://codeberg.org/ofosos/rotoscope version f80378ef7dbec85dd86b4427ca9e2eee5af60779

```bash
# tox
```

**Please provide the full traceback using the `--very-verbose` flag.**
```python
/Users/mark/PycharmProjects/rotoscope/venv/bin/Python /Applications/PyCharm.app/Contents/plugins/python/helpers/pycharm/_jb_tox_runner.py
Testing started at 23:10 ...
.package recreate: /Users/mark/PycharmProjects/rotoscope/.tox/.package
.package installdeps: setuptools>=46.1.0, setuptools_scm[toml]>=5, wheel
WARNING: Discarding $PYTHONPATH from environment, to override specify PYTHONPATH in 'passenv' in your configuration.
/Users/mark/PycharmProjects/rotoscope/.tox/.package/lib/python3.8/site-packages/setuptools_scm/version.py:67: UserWarning: tag 'draft1' no version found
  warnings.warn(f"tag {tag!r} no version found")
Traceback (most recent call last):
  File "venv/lib/python3.8/site-packages/tox/helper/build_requires.py", line 16, in <module>
    for_build_requires = backend.get_requires_for_build_sdist(None)
  File "/Users/mark/PycharmProjects/rotoscope/.tox/.package/lib/python3.8/site-packages/setuptools/build_meta.py", line 181, in get_requires_for_build_sdist
    return self._get_build_requires(config_settings, requirements=[])
  File "/Users/mark/PycharmProjects/rotoscope/.tox/.package/lib/python3.8/site-packages/setuptools/build_meta.py", line 159, in _get_build_requires
    self.run_setup()
  File "/Users/mark/PycharmProjects/rotoscope/.tox/.package/lib/python3.8/site-packages/setuptools/build_meta.py", line 174, in run_setup
    exec(compile(code, __file__, 'exec'), locals())
  File "setup.py", line 13, in <module>
    setup(use_scm_version={"version_scheme": "no-guess-dev"})
  File "/Users/mark/PycharmProjects/rotoscope/.tox/.package/lib/python3.8/site-packages/setuptools/__init__.py", line 87, in setup
    return distutils.core.setup(**attrs)
[...]

PR #651, opened 2022-06-18
Add step to migration docs alerting about tags

abravalheri:
## Purpose
In #649 we learn that `setuptools-scm` might have trouble with existing tags that don't resemble a version.

I don't know if it is intentional, but this is a confusing factor for new comers, specially if they want to migrate an existing repository to `pyscaffold`.

We should add least add some note about it. 

## Approach
- Add a note to the migration guides about verifying the tags.

## Resources & Links

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -394,8 +394,7 @@ How can I configure and debug the exact version exported by my package?
 Why do I see the error ``AssertionError: Can't parse version ...``?
     The way ``setuptools-scm`` parses git tags has some edge cases leading
     to this assertion error. The best way is to either remove the tag or
-    rename it to either a proper semantic version like ``v0.4.2`` or avoid
-    using any number (e.g. ``draft``).
+    rename it to a proper semantic version like ``v0.4.2``.
 
     Please also note that ``setuptools-scm`` may also fail if a single commit
     has multiple tags and at least one of them is problematic (``setuptools-scm``
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -391,7 +391,7 @@ How can I configure and debug the exact version exported by my package?
 
     (This requires the ``setuptools-scm`` package is installed in your environment)
 
-Why do I see the error ``AssertionError: cant parse version ...``?
+Why do I see the error ``AssertionError: Can't parse version ...``?
     There are some limitations with ``setuptools-scm`` and in some edge cases
     it might fail to parse specific tags.
     If that is happening to your project (other than removing the problematic tag),
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -391,6 +391,17 @@ How can I configure and debug the exact version exported by my package?
 
     (This requires the ``setuptools-scm`` package is installed in your environment)
 
+Why do I see the error ``AssertionError: cant parse version ...``?
+    There are some limitations with ``setuptools-scm`` and in some edge cases
+    it might fail to parse specific tags.
+    If that is happening to your project (other than removing the problematic tag),
+    you can try to re-run the relevant command after setting the
+    ``SETUPTOOLS_SCM_PRETEND_VERSION`` environment variable to overwrite the
+    default behavior.
+    Please also note that ``setuptools-scm`` may also fail if a single commit
+    has multiple tags and at least one of them is problematic (``setuptools-scm``
+    *will not* try multiple tags until it finds the correct one).
+
 
 .. _blog post by Ionel: https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure
 .. _src layout: https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -392,12 +392,11 @@ How can I configure and debug the exact version exported by my package?
     (This requires the ``setuptools-scm`` package is installed in your environment)
 
 Why do I see the error ``AssertionError: Can't parse version ...``?
-    There are some limitations with ``setuptools-scm`` and in some edge cases
-    it might fail to parse specific tags.
-    If that is happening to your project (other than removing the problematic tag),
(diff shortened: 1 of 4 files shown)


docs/faq.rst
```diff
@@ -391,6 +391,17 @@ How can I configure and debug the exact version exported by my package?
 
     (This requires the ``setuptools-scm`` package is installed in your environment)
 
+Why do I see the error ``AssertionError: cant parse version ...``?
```
FlorianWilhelm, line 394: The error is: `AssertionError: Can't parse version`, right?

docs/faq.rst
```diff
@@ -391,6 +391,17 @@ How can I configure and debug the exact version exported by my package?
...
+    There are some limitations with ``setuptools-scm`` and in some edge cases
+    it might fail to parse specific tags.
+    If that is happening to your project (other than removing the problematic tag),
```
FlorianWilhelm, line 397: I am not super sure about it, but should it rather say that `setuptools-scm` expects version tags according to the [semantic versioning scheme](https://semver.org/). Thus, rename your tags to look like `v1.3` or `v1.3.2`.

docs/faq.rst
```diff
@@ -391,6 +391,17 @@ How can I configure and debug the exact version exported by my package?
...
+    you can try to re-run the relevant command after setting the
+    ``SETUPTOOLS_SCM_PRETEND_VERSION`` environment variable to overwrite the
+    default behavior.
```
FlorianWilhelm, line 400: Giving here an example might help. So `export SETUPTOOLS_SCM_PRETEND_VERSION=v1.4` in Bash, right?

FlorianWilhelm (commented):
Thanks @abravalheri!

docs/faq.rst
abravalheri: Currently `setuptools-scm` seems to aim for ignoring tags that don't resemble a version tag.
For example:
- it would ignore a tag named `draft`
- it would parse a tag named `draft-1` and give back version `1` (it assumes that the tag is in the format `<package name>-<version number>`)

However there is an edge case which results in an exception when no `-` is used (e.g. `draft1`).

I noticed that `setuptools-scm` has an [issue open](https://github.com/pypa/setuptools_scm/issues/449) which seems to indicate that the problem is fixed. So I asked the developers about this behaviour in the discord channel: it might just be a limitation that `setuptools-scm` chooses to not handle, rather than an error.

What is weird to me is that the code path checks twice the tag format. The first time `setuptools-scm` sees the tag, it complains with a warning but keeps going on, the second time it raises an error via an assert statement.

That looks a bit odd. If the intention is to not support this scenario, it would seem better to just fail the first time it checks. Moreover, assert statements in production code are (most of the times) added when the developer already believes that the assertion will never fail and that all the edge cases have already been handled. But yeah, I don't know much about `setuptools-scm` code base so let's see if they can clarify that detail...

docs/faq.rst
abravalheri, line 400: Thank you, that is a good suggestion.

```suggestion
    default behavior.
    For example, if you are using bash in your terminal, you can run
    `export SETUPTOOLS_SCM_PRETEND_VERSION=v0.42`
    before running the tox commands.
```

This reminds me that we should add `SETUPTOOLS_SCM_PRETEND_VERSION` to the `passenv` configuration in the tox file to allow this workaround to work... What do you think?

abravalheri (commented):

I think I will leave this PR open at least for a bit to see if the `setuptools-scm` devs have any comments, but even if they fix the edge case in the associated issue, it might be relevant to add this to the FAQ (in the case there is another edge case we still don't know about it -- regexes are complicated 😅)

docs/faq.rst
abravalheri: For some reason, GitHub separated my reply in a different comment: https://github.com/pyscaffold/pyscaffold/pull/651#discussion_r901065806.

docs/faq.rst
FlorianWilhelm: Ah! Thanks for the explanation. So from a user's perspective, one could say:

> The way ``setuptools-scm`` parses git tags has some edge cases leading to this assertion error. The best way is to either remove the tag or rename it to either a proper semantic version like `v0.4.2` or a word without a number like `draft`.

docs/faq.rst
FlorianWilhelm: Cool idea!

docs/faq.rst
```diff
@@ -391,6 +391,16 @@ How can I configure and debug the exact version exported by my package?
 
     (This requires the ``setuptools-scm`` package is installed in your environment)
 
+Why do I see the error ``AssertionError: Can't parse version ...``?
+    The way ``setuptools-scm`` parses git tags has some edge cases leading
+    to this assertion error. The best way is to either remove the tag or
+    rename it to either a proper semantic version like ``v0.4.2`` or avoid
+    using any number (e.g. ``draft``).
```
abravalheri, line 398: Ronny confirmed in the PyPA discord chat that `setuptools-scm` has no intention of supporting non-version tags.

```suggestion
    rename it to a proper semantic version like ``v0.4.2``.
```

docs/migration.rst
```diff
@@ -39,6 +39,14 @@ Let's start:
    In most cases you will not need to make changes to the new ``setup.py`` file provided by PyScaffold.
    The only exceptions are if your project uses compiled resources, e.g. Cython.
 
+#. If you have any pre-existing `git tag`_ in your repository history, you will
+   need to ensure that the latest tag is compatible with :pypi:`setuptools-scm`.
+   You can do that by running ``python -m setuptools_scm`` (after installing it
+   in your environment). If the command succeeds you are good to go.
+
+   Please note that some very specific tag formats can be problematic.
```
abravalheri, line 47:
```suggestion
   Please note that some specific tag formats can be problematic.
```

PR #652, opened 2022-06-25
Remove redundant wheel dep from pyproject.toml

mgorny:
## Purpose
The `wheel` dependency in pyproject.toml is redundant, as it is added by the backend
automatically.  Listing it explicitly in the documentation was a historical mistake and has been fixed since.

## Approach
By removing it.

- [ ] Should I update the generated `pyproject.toml` scaffolds myself?

## Resources & Links

[https://github.com/pypa/setuptools/commit/f7d30a9529378cf69054b5176249e5457aaf640a](commit updating setuptools docs)

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -23,3 +23,4 @@ Contributors
 * Noah Pendleton
 * John Vandenberg
 * Ben Mares
+* Michał Górny
--- a/src/pyscaffold/dependencies.py
+++ b/src/pyscaffold/dependencies.py
@@ -8,7 +8,7 @@ from packaging.requirements import InvalidRequirement, Requirement
 
 # setuptools version is now enforced via `install_requires`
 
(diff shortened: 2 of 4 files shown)


abravalheri:
Thank you very much @mgorny .

abravalheri:
Regarding the question:

> Should I update the generated pyproject.toml scaffolds myself?

If you have already used PyScaffold to create a project, it should be completely fine to manually remove the wheel dependency from the `pyproject.toml` file.

mgorny:
> If you have already used PyScaffold to create a project, it should be completely fine to manually remove the wheel dependency from the `pyproject.toml` file.

I was asking about the `pyproject.toml` files inside the pyscaffold source tree.

abravalheri:
Aah,  I see. No problems, I can do that when preparing the release.

mgorny:
Ok, thanks a lot!

PR #655, opened 2022-07-08
Fix typos in installation instructions

cverrier:
## Purpose
There are some typos in `docs/install.rst` file.

## Approach
Fix the spotted typos in `docs/install.rst` file.


abravalheri:
Thank you very much @cverrier.

PR #674, opened 2022-10-19
Add codespell to pre-commit

abravalheri:
## Purpose
As a feature add `codespell` to the pre-commit config.

## Approach
This does not interfere with the current behaviour:
the new hook is not activated by default, the user has to opt-in by uncommenting it.


FlorianWilhelm:
Haven't heard of it but it looks cool! Thanks @abravalheri

PR #694, opened 2023-01-20
Add comments about  building `sdist` in `tox.ini`

abravalheri:
(*the tests are failing for reasons being tackled in #693*)

## Purpose
Some users may have problems with uploading `sdist` to PyPI (or they may not want to).

## Approach
This PR adds documentation about this problem:
- A comment on `tox.ini` about alternative usage of `tox -e build`.
- A section on FAQ about `sdists`.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -406,7 +406,7 @@ Common Doubts about ``sdist``
 
 Is ``sdist`` considered deprecated:
     Not really. Packaging in Python is done in two stages:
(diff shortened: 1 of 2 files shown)


FlorianWilhelm (approved):
Wow, thanks @abravalheri, as always, your work is very thorough and sophisticated. Sounds great and clears up things about sdists. I must say that I also learnt new things when reading this. Thank you!

PR #701, opened 2023-02-21
Make some ruff fixes

cclauss:
## Purpose
[`ruff --select=RET502,SIM117,SIM118,RUF005 --fix .`](https://beta.ruff.rs)
Executes the following fixes:
```
  3	RET502 	[*] Do not implicitly `return None` in function able to return non-`None` value
  1	SIM117 	[*] Use a single `with` statement with multiple contexts instead of nested `with` statements
  2	SIM118 	[*] Use `license in templates.licenses` instead of `license in templates.licenses.keys()`
  4	RUF005 	[*] Consider `["git", *list(args)]` instead of concatenation
```

## Approach

- [ ] Use GitHub checklists. When solved, check the box and explain the answer.

## Resources & Links

_Links to blog posts, patterns, libraries or add-ons used to solve this problem_

### DIFF
--- a/src/pyscaffold/file_system.py
+++ b/src/pyscaffold/file_system.py
@@ -270,7 +270,7 @@ def on_ro_error(func, path, exc_info):
     sleep(0.5)
 
     if not Path(path).exists():
-        return
+        return None
 
     if not os.access(path, os.W_OK):
         # Is the error an access error ?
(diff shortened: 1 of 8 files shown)


abravalheri:
Thank you very much @cclauss

PR #715, opened 2023-04-17
Changing `Github Token` variable in template.

victorgarcia98:
Hello!

Running `pyscaffold` with the the Github Actions flag I've encountered that It fails to upload the code coverage report. By default, GitHub exposes the *Github Token* through the variable `secrets.GITHUB_TOKEN`  in [upper case]((https://docs.github.com/en/actions/security-guides/automatic-token-authentication)), while in the template it is used in lower case `secrets.github_token`.

Thanks,
Víctor


abravalheri:
Thank you very much @victorgarcia98

PR #721, opened 2023-05-30
Allow platformdirs 3.x (fix #719)

### LINKED ISSUE #719: ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts. pyscaffold 4.4.1 requires platformdirs<3,>=2, but you have platformdirs 3.5.0 which is incompatible.

PyScaffold uses outdated `platformdirs`

1. Install pyscaffold
```bash
pip install --upgrade pyscaffold
```

2. Install `tox`
 ```bash
pip install tox
```

**Please provide the full traceback using the `--very-verbose` flag.**
```python
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
pyscaffold 4.4.1 requires platformdirs<3,>=2, but you have platformdirs 3.5.0 which is incompatible
```

This occurs because `tox` requires `platformdirs>3.2`.

* PyScaffold Version: 4.4.1
* Python Version: 3.9.6
* Operating system: MacOS
* How did you install PyScaffold: pip

musicinmybrain:
## Purpose

This PR expands the allowed versions of the `platformdirs` dependency from 2.x only to 2.x *or* 3.x. This would fix https://github.com/pyscaffold/pyscaffold/issues/719 and is required downstream in Fedora Linux for compatibility with the current version of `platformdirs`.

## Approach

The upper version bound in `setup.cfg` is trivially adjusted.

[x] Ran `tox` locally with no regressions

## Resources & Links

The sole breaking change in the [changelog for `platformdirs` 3.0.0](https://github.com/platformdirs/platformdirs/blob/3.5.1/CHANGES.rst#platformdirs-300-2023-02-06) is:

- **BREAKING** Changed the config directory on macOS to point to ``*/Library/Application Support``

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,18 @@ Changelog
     -------------------------
 
 
+Development version
+===================
+
+Version 4.5.0, 2023-XX-XX
+-------------------------
+
+- Change macOS default configuration dir from
+  ``~/Library/Preferences`` to ``~/Library/Application Support``.
+  This change is motivated by a change in the ``platformdirs`` dependency.
+  You can read more about the motivation in :github:`platformdirs/platformdirs#98`
+
+
 Current versions
 ================
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,7 +18,7 @@ Version 4.5.0, 2023-XX-XX
 -------------------------
 
 - Change macOS default configuration dir from
-  ``~/Library/Preferences`` to ``~/Library/Application Support``.
+  ``~/Library/Preferences`` to ``~/Library/Application Support``, :pr:`721`.
   This change is motivated by a change in the ``platformdirs`` dependency.
   You can read more about the motivation in :github:`platformdirs/platformdirs#98`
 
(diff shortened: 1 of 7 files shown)


abravalheri:
Thank you very much for the contribution @musicinmybrain.

> The sole breaking change in the [changelog for platformdirs 3.0.0](https://github.com/platformdirs/platformdirs/blob/3.5.1/CHANGES.rst#platformdirs-300-2023-02-06) is:
> BREAKING Changed the config directory on macOS to point to `*/Library/Application Support`

This might mean that existing configuration files for PyScaffold in macOS will stop working because they will no longer be in the right place, right?

---

On a unrelated note, @FlorianWilhelm, it seems that the tests started failing recently on macOS with a weird error message: `locale.Error: unsupported locale setting`.

It seems that this is related to a recent change in `sphinx`: https://github.com/pyscaffold/pyscaffold/pull/717#issuecomment-1540153892. Do you have any idea on how to deal with that?

abravalheri:
I added automatic migration to ensure backward compatibility.

musicinmybrain:
> I added automatic migration to ensure backward compatibility.

Thanks!

Issue #727, opened 2023-06-28
Should `.readthedocs.yml` be `.readthedocs.yaml` instead?


sgbaird:
## Description of your problem
The content looks like it's been updated to the new configuration format (https://github.com/pyscaffold/pyscaffold/blob/master/src/pyscaffold/templates/rtd_cfg.template); the file just isn't named what I expected based on the docs. I recently ran into issues with building on sphinx due to using the old file format. Maybe it doesn't matter to readthedocs if it's `.yml` vs. `.yaml`, but I figured I'd bring it up.

**Please provide a minimal, self-contained, and reproducible example.**
https://colab.research.google.com/gist/sgbaird/6e503e7f069260b039d6e86aeaf62721/pyscaffold-rtd-cfg-file.ipynb

**Please provide the full traceback using the `--very-verbose` flag.**
N/A

## Versions and main components

* PyScaffold Version: 4.5
* Python Version: 3.10.12
* Operating system: Ubuntu 20.04.6 LTS
* How did you install PyScaffold: (conda/**pip**)

PR #728, opened 2023-06-28
rename `.readthedocs.yml` to `.readthedocs.yaml`

sgbaird:
The content looks like it's been updated to the new configuration format (https://github.com/pyscaffold/pyscaffold/blob/master/src/pyscaffold/templates/rtd_cfg.template); the file just isn't named what I expected based on the docs. I recently ran into issues with building on sphinx due to using the old file format. Maybe it doesn't matter to readthedocs if it's `.yml` vs. `.yaml`, but I figured  I'd bring it up.

https://colab.research.google.com/gist/sgbaird/6e503e7f069260b039d6e86aeaf62721/pyscaffold-rtd-cfg-file.ipynb

## Purpose
Rename `.readthedocs.yml` to `.readthedocs.yaml` to be consistent with the readthedocs docs https://docs.readthedocs.io/en/stable/config-file/v2.html

image

https://github.com/pyscaffold/pyscaffold/issues/727

## Approach
Rename file

### DIFF
--- a/src/pyscaffold/structure.py
+++ b/src/pyscaffold/structure.py
@@ -126,7 +126,7 @@ def define_structure(struct: Structure, opts: ScaffoldOpts) -> ActionParams:
         # Tools
         ".gitignore": (get_template("gitignore"), NO_OVERWRITE),
         ".coveragerc": (get_template("coveragerc"), NO_OVERWRITE),
-        ".readthedocs.yml": (get_template("rtd_cfg"), NO_OVERWRITE),
+        ".readthedocs.yaml": (get_template("rtd_cfg"), NO_OVERWRITE),
         # Project configuration
         "pyproject.toml": (templates.pyproject_toml, NO_OVERWRITE),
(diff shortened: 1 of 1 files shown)


abravalheri:
Hi @sgbaird, thank you very much for opening this issue and providing a PR.
Can you describe the errors you were seeing?

In principle it should not matter if we use `.yml` or `.yaml` based on [these docs](https://dev.readthedocs.io/en/latest/design/yaml-file.html#format) (also the RTD documentation itself seems to use a [`.yml`](https://github.com/readthedocs/readthedocs.org/blob/9.16.0/.readthedocs.yml)), but there is no problem in changing it to match what is documented. I just want to make sure there is nothing else we need to change.

sgbaird:
@abravalheri you're right, shouldn't be an issue either way. I think `.readthedocs.yaml` can get confused with the prior format (it did for me at least), but not a major issue.

sgbaird:
I can close if you want.

PR #735, opened 2023-10-24
Modify Cirrus CI config to save resources

abravalheri:
## Purpose
Recently Cirrus CI changed its free limits for open source. The idea behind this PR is to minimise the usage via a "fail fast" approach.

## Approach

The changes can be summarised as follows:

1. Run "build" and "check" inside the same container, so we save the time to spawn a separated container.
2. Decrease the amount of Linux containers used. I left the lower and upper limits for the Python version in PR builds and added an intermediary version to the master/main branch.
3. Wait until the Linux tests pass before running more expensive ones (Windows, OS X, FreeBSD).

## Resources & Links

https://cirrus-ci.org/blog/2023/07/17/limiting-free-usage-of-cirrus-ci/

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -62,9 +62,9 @@ env:
 # ---- CI Pipeline ----
 
 build_task:
-  name: build (Linux - 3.10)
+  name: build and check (Linux - 3.11)
   alias: build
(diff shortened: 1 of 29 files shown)


abravalheri:
There are other errors still happening, but it is probably better to split the PRs

PR #742, opened 2023-12-20
Use version instead of  mutable reference for black

stijnvanhoey:
## Purpose
A pre-commit autoupdate will be needed to make sure latest versions are used, but this commit just overcomes the `[WARNING] The 'rev' field of repo 'https://github.com/psf/black' appears to be a mutable reference...` message when running the pre-commit after project setup.

## Approach
By providing a version tag instead of `stable`, the `mutable reference.` is removed.

### DIFF
--- a/src/pyscaffold/templates/pre-commit-config.template
+++ b/src/pyscaffold/templates/pre-commit-config.template
@@ -41,7 +41,7 @@ repos:
   - id: isort
 
 - repo: https://github.com/psf/black
-  rev: stable
+  rev: 23.12.0
   hooks:
   - id: black
     language_version: python3


abravalheri:
Thank you very much @stijnvanhoey.

Because, `black` constantly updates its version, I don't think it makes sense for us to continuously update this reference (cat-and-mopuse).

It is probably more convenient if the users run `pre-commit autoupdate`.

stijnvanhoey:
> Because, `black` constantly updates its version, I don't think it makes sense for us to continuously update this reference (cat-and-mouse game).

That is correct and my PR was not intended in continuously updating the black version. By providing just 'any' version instead of `rev: stable`, the warning about `mutable reference` is no longer returned when running the autoupdate. 

> It is probably more convenient if the users run `pre-commit autoupdate` (which I suppose is an instruction that also appears in the text displayed by `pre-commit`, correct?).

Indeed, the suggestion to run `pre-commit autoupdate` is still valid and stays the same. This Pr actually solves an issue when users run the command. When users run the autoupdate command, they will currently get the warning while this PR removes the warning when running the autoupdate command.

PR #747, opened 2024-03-28
Fall back to pkgutil when Python >= 3.13

penguinpee:
The `read_text` and other methods will be removed from Python 3.13.
As a stop gap measure  fall back to `pkgutil`.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -33,7 +33,7 @@ classifiers =
 [options]
 zip_safe = False
 packages = find_namespace:
-python_requires = >=3.6
+python_requires = >=3.7
 include_package_data = True
 package_dir =
     =src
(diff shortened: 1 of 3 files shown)


penguinpee:
Apologies for the somewhat messy PR. I tried a different approach, first. But that didn't work and I subsequently forgot to change the title of the commit message.

This might not be making it as the final solution. But it does solve the issue at our end regarding removal of `read_text`.

See also: https://docs.python.org/3.13/whatsnew/3.13.html#importlib

abravalheri:
Thank you very much @penguinpee for the contribution, and sorry for the time it took for the review.

We probably can use a different approach using `importlib.resources.files` (I think the core developers pretty much also want to remove most of `pkgutil` in the long term). I plan to push that solution as a follow up commit in this same PR (we will probably have to drop 3.6, but it is been a while that it reached its EoL).

abravalheri:
I will merge this now and leave the other CI problems for a follow up PR.

Thank you very much!

PR #757, opened 2024-10-07
Install missing template files

sunpoet:
## Purpose

FreeBSD ports uses sdist to build python modules. In this way, the template files are missing, therefore the following error occurs.

```
% putup
ERROR: There was an error loading 'pyscaffold.extensions.interactive'.
    Please make sure you have installed a version of the extension that is compatible
    with PyScaffold 4.6. You can also try unininstalling it.
```

```
% python3.11
Python 3.11.10 (main, Oct  7 2024, 15:23:45) [Clang 17.0.6 (https://github.com/llvm/llvm-project.git llvmorg-17.0.6-0-g600970 on freebsd13 Type "help", "copyright", "credits" or "license" for more information.
>>> from pyscaffold.extensions import interactive
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.11/site-packages/pyscaffold/extensions/interactive.py", line 42, in <module>
    HEADER = templates.get_template("header_interactive")
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/pyscaffold/templates/__init__.py", line 121, in get_template
    data = read_text(relative_to, file_name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/pyscaffold/templates/__init__.py", line 26, in read_text
    return files(package).joinpath(resource).read_text(encoding="utf-8")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/pathlib.py", line 1058, in read_text
    with self.open(mode='r', encoding=encoding, errors=errors) as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/pathlib.py", line 1044, in open
    return io.open(self, mode, buffering, encoding, errors, newline)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/usr/local/lib/python3.11/site-packages/pyscaffold/templates/header_interactive.template'
```

## Approach

This patch installs the missing template files and fix the module.

## Resources & Links

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=281745
https://cgit.freebsd.org/ports/commit/?id=0d231ec121aee06a694a5024fafc35622d90263d

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -55,6 +55,10 @@ install_requires =
 # However we specify a higher version so we encourage users to update the
 # version they have installed...
 
+[options.package_data]
+pyscaffold.templates =
+	*.template
+
 [options.packages.find]
 where = src
 exclude =


abravalheri:
Hi @sunpoet, thank you very much for reporting this. It is interesting that you are hitting that problem.

In principle, `pyscaffold` uses a combination of `setuptools-scm` (which is supposed to add ***all*** tracked files to the sdist) and `include_package_data=True` (which is supposed to add all files that co-inhabit the package folders to the wheel). So it is designed to intentionally to ditch the painful enumeration via the `package_data` configuration, and in theory the change you are suggesting is unnecessary.

We can see this mechanism working on the packages submitted to PyPI. For example, the `header_interactive.template` file is present in both

- `sdist`:  https://inspector.pypi.io/project/pyscaffold/4.6/packages/a2/97/2b2762142ff73ee1e3d6452c2fe0149dd0da36bf5f042e952c460e09ab74/pyscaffold-4.6.tar.gz/pyscaffold-4.6/src/pyscaffold/templates/header_interactive.template and
- `wheel`: https://inspector.pypi.io/project/pyscaffold/4.6/packages/b2/5e/3afc55a61171ab0ab8f72478d2732cf89f2a86865bce231d1e46e4191d78/PyScaffold-4.6-py3-none-any.whl/pyscaffold/templates/header_interactive.template

So I wonder if this, rather than a "misconfiguration" problem, this is actually caused by some particular way that these packages are being built that deviates from the way they are maintained in this repo?

abravalheri:
What happens if we bump the build dependencies on `setuptools` and `setuptools-scm` to more recent releases? Maybe that is the problem?

sunpoet:
Sorry for the noise. After further investigation, I noticed that it was caused by [the local change](https://github.com/freebsd/freebsd-ports/blob/main/devel/py-pyscaffold/files/patch-setup.cfg#L5-L6) of using lowercase name. The problem should not happen to other users. Therefore, it should still be a loca patch.

Issue #761, opened 2025-10-16
Minor language errors have been found in the templates.


gchang12:
## Description of your problem

**Please provide a minimal, self-contained, and reproducible example.**
```python
#
```

**Please provide the full traceback using the `--very-verbose` flag.**
```python
#
```

**Please provide any additional information below.**

## Versions and main components

* PyScaffold Version:
* Python Version:
* Operating system:
* How did you install PyScaffold: (conda/pip)

PanPriyanshu:
Hi! I’m interested in contributing to PyScaffold and preparing for GSoC 2026.
Could I work on this issue? I can start by reviewing the templates and fixing the minor language/grammar errors.

PR #762, opened 2025-10-16
Fix language in templates

gchang12:
## Purpose
_There are minor language issues in the templates. The purpose of this series of commits is to fix those._
https://github.com/pyscaffold/pyscaffold/issues/761

## Approach
_Manually corrected errors._

- [check] 'note', not 'notice'
- [check] 'look at', not 'look on'

## Resources & Links

### DIFF
--- a/src/pyscaffold/templates/sphinx_conf.template
+++ b/src/pyscaffold/templates/sphinx_conf.template
@@ -15,9 +15,9 @@ import shutil
 
 __location__ = os.path.dirname(__file__)
 
-# If extensions (or modules to document with autodoc) are in another directory,
-# add these directories to sys.path here. If the directory is relative to the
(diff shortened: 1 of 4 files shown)


src/pyscaffold/templates/sphinx_conf.template
```diff
@@ -145,7 +145,7 @@ pygments_style = "sphinx"
 # If true, keep warnings as "system message" paragraphs in the built documents.
 # keep_warnings = False
 
-# If this is True, todo emits a warning for each TODO entries. The default is False.
+# If this is True, a warning is emitted for each TODO entries. The default is False.
 todo_emit_warnings = True
```
gchang12: Also, note that `todo_emit_warnings` is by default equal to `True`, whereas the accompanying in-line comment says that it's equal to `False`.

